PR #15, opened 2016-01-20
Update repo URL

jubalh:
Old adress still works because it forwards to the new one. Anyways I
think the correct one should be used.


hennevogel:
cool! :sunglasses:

PR #20, opened 2016-04-27
Fix some minor issues

bgeuken:


xml/obs_best_practice_local_setup.xml
```diff
@@ -153,9 +153,9 @@
    need to work on several further steps. Open
    /usr/share/doc/packages/obs-api/README.SETUP file. To run OBS server, you
    need to follow each step of README.SETUP carefully based on your system
-   situation. You could also find this <filename>README.SETUP</filename>
-   file online using this URL
-   <ulink url="http://gitorious.org/opensuse/build-service/blobs/raw/master/dist/README.SETUP">http://gitorious.org/opensuse/build-service/blobs/raw/master/dist/README.SETUP</ulink>.
+   situation. You can also find this <filename>README.SETUP</filename>
+   file in our public 
+   <ulink url="https://github.com/openSUSE/open-build-service/blob/master/dist/README.SETUP.md">git repository</ulink>.
```
adrianschroeter: Maybe we should mention that this is the unstable version and people should check for the right branch?

Maybe we should create a "stable" branch which we can switch on new releases to keep such links static .....

bgeuken: Oh, right. Didn't think of that

bgeuken: The paragraph now mentions the stable branches, https://github.com/openSUSE/obs-docu/pull/20/commits/0021cee5799ba70953d737af6d4a1316bd53946f#diff-9b211b9bf87be0f33a22f541ce92bcb3R159

PR #38, opened 2016-11-15
Add documentation for spider identification

mschnitzer:
This adds documentation for the new spider identification process.

See the attached PDF file for the whole documentation.
[obs-admin-guide_color_en.pdf](https://github.com/openSUSE/obs-docu/files/591177/obs-admin-guide_color_en.pdf)

PR #40, opened 2016-11-24
Spider identification

bgeuken:


bgeuken:
Updated according to https://github.com/openSUSE/obs-docu/pull/38#discussion_r89491970

@hennevogel please review

xml/obs_ag_spider_identification.xml
```diff
@@ -0,0 +1,73 @@
...
+  "http://www.docbook.org/xml/4.5/docbookx.dtd"
+[
+  <!ENTITY % entities SYSTEM "entity-decl.ent">
+  %entities;
+]>
+<section id="_obs_spider_identification">
+ <title>Spider Identification</title>
+ <para>
+   OBS is hiding specific parts/pages of the application from search crawlers (duckduckgo, google etc.), mostly for performance reasons.
+   Which user-agent strings are identified as crawlers is configured by the <ulink url="https://github.com/monperrus/crawler-user-agents/blob/master/crawler-user-agents.json">upstream crawler-user-agents.json</ulink> file.
```
hennevogel: Why is an upstream file? It's not created by default in our code base? It should be I guess...

bgeuken: Not sure if it's worth to explicitly create that file. On the other hand, it would make it more intuitive for OBS admins.

hennevogel: Yeah I guess the gem ships it. But I would agree, make this easy for OBS admins, those are the people we do this for...

bgeuken:
@hennevogel Update docu. PR for OBS follows soon.

PR #43, opened 2017-01-17
Mention admin guide in README

bgeuken:


mschnitzer:
lgtm

README
```diff
@@ -7,6 +7,7 @@ the Open Build Service (OBS).
 
 The documentation is currently split into two books:
```
eduardoj: two -> three. Isn't it?

bgeuken: Good catch

bgeuken:
\*Updated\*

PR #53, opened 2017-02-08
Clarify section regarding auth token generation and execution

asmorodskyi:


xml/obs_authorization_token.xml
```diff
@@ -68,23 +68,31 @@
...
+  <para>
+  The best way to create a token is bind it to a specific package. The advantage is that
+  the operation is limited to that package, so less bad things can happen
+  when the token leaks.
+  </para>
+<screen><command>osc token --create &lt;PROJECT&gt; &lt;PACKAGE&gt;</command></screen>
+<para>
+  Also, you do not need to specify the package at execution time.
+  But keep in mind that such form only works when you run it on an as
+  checkout of a package. Both commands below do the same thing in a different way:
```
adrianschroeter: It would be a bug if that command is not running without checkout. it works for me also without one ...

asmorodskyi: Didn't get what I need to change here. Can you please clarify ?

adrianschroeter: You can run eg. "osc token --create home:adrianSuSE uranium" without any checkout, but I read your text that you need a checked out package to create a package specific token.

asmorodskyi: "Also, you do not need to specify the package at execution time."  - command that you provide conflicts with this statement. I describing totally different scenario here. When inside a package folder you can run command without specifying a package

PR #59, opened 2017-03-09
Initail scratch service example

k0da:
Signed-off-by: Dinar Valeev <k0da@opensuse.org>

### DIFF
--- a/xml/obs_source_service.xml
+++ b/xml/obs_source_service.xml
@@ -203,7 +203,17 @@
  <section>
   <title>How to write a source service</title>
   <para>
-   ..
+    The logic behind it is an XML formated recipe which describes command and its parameters, like:
+    <screen><command>
+    &lt;services&gt;
+      &lt;service name="my_script_to_execute"&gt;
+      &lt;param name="param1"&gt;argument_value&lt;/param&gt;
+      &lt;/service&gt;
+      &lt;/services&gt;
+      the above example will execute <command>/usr/lib/obs/service/my_script_to_execute --param1 argument_value --outdir curr_dir</command>
+
+      It is important to mention that --outdir argument is a mandatory one and should be handled in each service which should
+      point to directory where the result will be picked from later by osc(localrun) or sourceserver(remoterun)
   </para>
  </section>
  <section>


xml/obs_source_service.xml
```diff
@@ -203,7 +203,17 @@
...
+    <screen><command>
+    &lt;services&gt;
+      &lt;service name="my_script_to_execute"&gt;
+      &lt;param name="param1"&gt;argument_value&lt;/param&gt;
+      &lt;/service&gt;
+      &lt;/services&gt;
+      the above example will execute <command>/usr/lib/obs/service/my_script_to_execute --param1 argument_value --outdir curr_dir</command>
+
+      It is important to mention that --outdir argument is a mandatory one and should be handled in each service which should
+      point to directory where the result will be picked from later by osc(localrun) or sourceserver(remoterun)
```
fsundermeyer: There are a few syntax errors here and I would also suggest some minor content adjustments:

    <para>
     To define a source service add a <sgmltag>service</sgmltag> section to the <filename>_service</filename> file of the project. Within this section, define a script and its arguments that will be called by the source service. The script needs to reside in <filename>/usr/lib/obs/service</filename>.
    </para>
    <screen>&lt;services&gt;
     &lt;service name="<replaceable>MY_SCRIPT</replaceable>"&gt;
      &lt;param name="<replaceable>PARAMETER1</replaceable>"&gt;<replaceable>PARAMETER1_VALUE<replaceable>&lt;/param&gt;
     &lt;/service&gt;
    &lt;/services&gt;</screen>
    <para>
     The example above will execute the following command:
    </para>
    <screen>/usr/lib/obs/service/ <replaceable>MY_SCRIPT</replaceable> \
    --<replaceable>PARAMETER1</replaceable> <replaceable>PARAMETER1_VALUE<replaceable> --outdir <replaceable>DIR</replaceable></screen>
    <para>
     The <option>--outdir</option> parameter is mandatory and needs to be handled in each service script. It needs to point to a directory where the result will be picked up by <command>osc</command> (localrun) or <command>sourceserver</command> (remoterun).
    </para>

The last paragraph is a bit unclear to me:
* where is the value for --outdir defined? It does not show up in the XML
* is it handled by the script (as I assumed)
* Are osc(localrun) and sourceserver(remoterun) explained anywhere else? If so, it would be good to add a pointer to the respective sections (&lt;xref linkend="ID"/&gt;)

adrianschroeter: --outdir can't be defined by the user, it is just required by our tool chain. It must not show up in the XML

Yes, this the script needs to accept it at least. And store modified data there.

tomschr:
Thanks Dinar for your contribution! 👍

The guides are currently reorganized, so I've integrated your text into the `develop` branch now at commit c16bb41c89 and closing this PR. 
If you would like to further amend this, please use it from the `develop` branch. Once it is finished, all changes will be integrated into `master` again.

PR #61, opened 2017-04-21
Add a note on handling admin accounts in kerberos mode

bgeuken:


bgeuken:
@eduardoj Please review

Issue #3, opened 2013-10-11
Reference Guide is incomplete on many places


okin:
The reference guide is incomplete at many, many places and thus it is not a very good reference for someone seeking to learn more about the Open Build Service.

PR #62, opened 2017-06-02
Remove all empty documentation sections

okurz:
They are just confusing and disappointing users and provide a false impression.

Less is more :-)

### DIFF
--- a/xml/factory.xml
+++ b/xml/factory.xml
@@ -18,7 +18,7 @@
    The main project is openSUSE:Factory. This project is controlled by a small
    group which does review all submissions according to the policies.
    Submissions are possible via submit requests, which get reviewed by default
-   by two groups: The Legal team and the code review team. ...
+   by two groups: The Legal team and the code review team.
   </para>
  </section>
  <section id="cha.obs.factory.develprojects">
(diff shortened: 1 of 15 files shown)


okurz:
Fixing https://github.com/openSUSE/obs-docu/issues/3

hennevogel:
So hiding titles of non existing topics helps people looking for documentation on those topics how?

okurz:
* saves time not wasted on looking for an empty section
* less dissappointment -> * less grumpy comments by people on IRC that the documentation is incomplete

hennevogel:
People are rightfully grumpy. And nothing will be better for them if we remove the empty sections, so I they will still complain. I don't follow your argument, sorry

okurz:
Complaining about "documentation could be better" is one thing - of course it can always be better. But having empty sections for so many years is just misleading. People get their hopes up and then are disappointed. Why not keep it simple and the documentation mentions everything that is covered and that's it. If I would understand more about OBS I would maybe contribute to it but probably not even to the empty sections because I don't know what the original author had as intentions. I think every other potential contributor would feel the same. Keeping sections with "to be implemented…" is like the webpages from the 1990s when so many people wanted to jump the train - "I can have my own webpage, yay!" - but did not know what to put there so you are greeted with these animated construction site gif's which made your eyes bleed and you never take a look again on that page - https://motherboard.vice.com/en_us/article/this-guy-compiled-every-under-construction-gif-from-the-90s

Did I do anything wrong from technical perspective, e.g. regarding the syntax? Did I delete wrong tags or should I delete more to make it complete?

PR #69, opened 2017-08-16
Nitpickery in beginner's guide

sknorr:
@tomschr have fun ... sorry, this has become so much. lots of those are just language issues which you don't need to look at closely. but there's a good number of remarks too.

in cases where i removed my own comments, i have seen the light and wholeheartedly agree with you now.

### DIFF
--- a/xml/art-obs-beginners-guide.xml
+++ b/xml/art-obs-beginners-guide.xml
@@ -31,8 +31,6 @@ https://www.suse.com/communities/blog/suse-studio-integration/
    </para>
   </abstract>
  </info>
- <remark>toms 2017-08-08: Generally, describe some/all use cases additionally
-  with the Web UI?</remark>
 
  <sect1 xml:id="sec.obsbg.targetaudience">
   <title>Target Audience</title>
@@ -77,7 +75,7 @@ https://www.suse.com/communities/blog/suse-studio-integration/
        Packages depend on other packages to function properly.
        There are two types of requirements:
        <emphasis>build requirements</emphasis> and
-       <emphasis>install requirements</emphasis>.
+       <emphasis>installation requirements</emphasis>.
       </para>
      </formalpara>
      <para>
@@ -85,13 +83,8 @@ https://www.suse.com/communities/blog/suse-studio-integration/
       build process in &obsa;. For example, a C++ program needs
       a C++ compiler.
      </para>
-     <remark>
-      "Install requirement" seems like bad English. Additionally, there is
-      the the distinction between run-time deps and actual install pre-reqs.
-      - sknorr, 2017-08-15
-     </remark>
      <para>
-      Install requirements are dependencies which are needed when installing
+      Installation requirements are dependencies which are needed when installing
       the final package.
      </para>
     </listitem>
@@ -112,18 +105,12 @@ https://www.suse.com/communities/blog/suse-studio-integration/
     in a file with the file extension <filename
      class="extension">.spec</filename>.
    </para>
-   <remark>toms 2017-08-08: Describe other naming conventions which applies only
-   for certain distributions</remark>
-   <remark>toms 2017-08-08: Describe Debian rule files?</remark>
   </sect2>
   <sect2 xml:id="sec.obsbg.buildhost">
    <title>Build Hosts and Packages</title>
   <para>
-   The &obs; is accessible either through a Web interface or a command line
-   interface (&osccmd;). Both are clients for the &obsa; API which provides
-   access to the database,
-   the search engine, and more. The OBS API also indirectly communicates with
-   the back-end which provides a storage pool and the build hosts.
+   The &obsa; server provides a Web interface and an API. The API is
+   used by the &osccmd; command line tool.
   </para>
   <para>
    To build the package, the back-end creates a sandbox with
@@ -173,7 +160,11 @@ https://www.suse.com/communities/blog/suse-studio-integration/
     the project level.
    </para>
    <para>
-    Projects can also contain other projects (subprojects) to structure work.
+    Projects can also have other projects (subprojects) to structure work.
+    They are isolated from their parents project and can be configured
+    independently.
+   </para>
+   <para>
     Each project name is separated by colons.
     For example, in the &opensuse; Build Service, packages for fonts are
     collected in the project
@@ -188,7 +179,10 @@ https://www.suse.com/communities/blog/suse-studio-integration/
     <firstterm baseform="Home Project">home project</firstterm>, available
     in &obsa; as <uri>home:<replaceable>USERNAME</replaceable></uri>.
     Home projects serve as a personal working area in &obsa; to define build
-    targets, upload, build, and download packages.
+    targets, upload, build, and download packages. Users are also permitted
+    to create subprojects f 
+    temporary
+    subprojects to work on other people's packages. 
    </para>
    <para>
     Sometimes, you will see the <uri>obs://<replaceable
@@ -218,8 +212,8 @@ https://www.suse.com/communities/blog/suse-studio-integration/
      <itemizedlist>
       <listitem>
        <para>
-        For &suse; related systems, install it with the
-        <command>zypper</command> command (replace
+        For &suse; related systems, install the <package>osc</package>
+        package with the <command>zypper</command> command (replace
         <replaceable>DISTRIBUTION</replaceable> with your distribution):
        </para>
        <screen>&prompt.root;<command>zypper</command> ar &obsrepourl;/openSUSE:/Tools/<replaceable>DISTRIBUTION</replaceable>/openSUSE:Tools.repo
@@ -260,63 +254,27 @@ https://www.suse.com/communities/blog/suse-studio-integration/
   <para>
    You can use the <command>osc</command> command without configuring your
    system as described in this section. However, we recommend preparing
-   your system for the following reasons:
+   your system with <command>sudo</command>.
   </para>
-  <itemizedlist>
-   <listitem>
-    <remark>
-     While the sudo setting really makes sense, I don't think the subdir
-     setting is important enough to be mentioned here... (I am not using it,
-     for example). - sknorr, 2017-08-15
-    </remark>
-    <remark>toms 2017-08-16: my consideration was that with subdirs enabled,
-     the user can't mix it up with a repo/project notation. Without the subdirs
-     enabled, the user could be confused by "cd into home:bla" instead of
-     "cd into home/bla".</remark>
-    <formalpara>
-     <title>Separate Projects in Subdirectories</title>
-     <para>
-      This is an optional configuration and a matter of taste. Whenever you
-      check out a project from your &obsa; instance, projects with colons are
-      not created as one single directory, but as subdirectories.
-     </para>
-    </formalpara>
-   </listitem>
-   <listitem>
-    <formalpara>
-     <title>Sudo Configuration</title>
-     <para>Although you can use &osccmd; as a normal user, at some point, you
-      need to enter the &rootuser; password. This can be annoying if you
-      rebuild packages frequently. Therefore, the configuration below customizes
-      the <command>sudo</command> configuration to allow building packages
-      without entering the &rootuser; password. To maximize security, only
-      specific users are allowed.
-     </para>
-    </formalpara>
-   </listitem>
-  </itemizedlist>
+  <para>
+   At some point, you need to enter the &rootuser; password. This can be
+   annoying if you rebuild packages frequently. Therefore, the configuration
+   below customizes
+   the <command>sudo</command> configuration to allow building packages
+   without entering the &rootuser; password. To maximize security, only
+   specific users are allowed.
+  </para>
 
-  <para>To use the above customization, execute the following steps:</para>
+  <para>
+   To customize <command>sudo</command>, execute the following steps:
+  </para>
   <procedure xml:id="pro.obsbg.obsconfig">
-   <title>Configuring <command>sudo</command> and <command>osc</command></title>
-   <step>
-    <remark>
-     An account is mentioned among the prereqs. You could move the prereq
-     section up and could then remove this first step...
-     - sknorr, 2017-08-15
-    </remark>
-    <para>
-     If you do not have already an account on your &obs; instance, create one.
-    </para>
-   </step>
-   <step>
-    <para>Configure <command>sudo</command>.</para>
-    <para>
+   <title>Configuring <command>sudo</command></title>
+   <para>
      The following configuration allows all users which belong to the group
      <systemitem class="groupname">osc</systemitem> to build packages without
-     entering the &rootuser; password:
+     entering the &rootuser; password.
     </para>
-    <substeps>
      <step>
       <para>Log in as &rootuser; and create a new group <systemitem
        class="groupname">osc</systemitem>.
@@ -349,13 +307,6 @@ Cmnd_Alias  OSC_CMD = /usr/bin/osc, /usr/bin/build
        effective.
       </para>
      </step>
-    </substeps>
-    <para>
-     Whenever you want to add additional users later, use the
-     <command>usermod</command> command as shown in
-     <xref linkend="st.oscbg.usermod"/>.
-    </para>
-   </step>
    <step>
     <para>Create a new &obsa; configuration file:</para>
     <screen>&prompt.user;<command>osc</command> ls &obshome1;</screen>
@@ -364,21 +315,22 @@ Cmnd_Alias  OSC_CMD = /usr/bin/osc, /usr/bin/build
        for your &obsa; user name and your &obsa; password.
       </para>
    </step>
-   <step xml:id="st.obsbg.obsconfig.nocolon">
-      <para>
-       Optionally, if you prefer to separate projects and subprojects
-       in directories and subdirectories, change the following line
-       in your configuration file <filename>~/.oscrc</filename>: 
-      </para>
-      <screen>checkout_no_colon = 1</screen>
-      <para>
-       This will use an alternate layout when checking out a project.
-       For example, setting the option above and checking out the home project
-       will lead to a directory structure <filename>&obsworkdir1;</filename>
-       instead of the single directory <filename>&obshome1;</filename>.
-      </para>
-     </step>
   </procedure>
+  
+  <para>
+   Optionally, if you prefer to separate projects and subprojects in
+   directories and subdirectories, change the following line in your
+   configuration file <filename>~/.oscrc</filename>: </para>
+  <screen>checkout_no_colon = 1</screen>
+  <para>
+   This will use an alternate layout when checking out a project. For
+   example, setting the option above and checking out the home project will
+   lead to a directory structure <filename>home/&obsuser1;</filename>
+   instead of the single directory <filename>&obshome1;</filename>. 
+  </para>
+  <para>
+   In this guide, the default configuration with colons is used.
+  </para>
  </sect1>
 
  <sect1 xml:id="sec.obsbg.usagescenarios">
@@ -388,9 +340,24 @@ Cmnd_Alias  OSC_CMD = /usr/bin/osc, /usr/bin/build
   </para>
   <itemizedlist>
    <listitem>
-    <para>You must have an account on an &obs; instance (for example, at
-     <link xlink:href="https://build.opensuse.org"/>).</para>
+    <para>
+     You are using the &osbs; at <link xlink:href="https://build.opensuse.org"/>.
+     If you are using another &obsa; instance, some commands may differ.
+    </para>
    </listitem>
+   <listitem>
+    <para>You have an account on an &obs; instance.</para>
+   </listitem>
+   <listitem>
+    <para>An RPM-based operating system like &opensuse; or &sle;.</para>
+   </listitem>
+   <listitem>
+    <para>A customized system as shown in <xref
+     linkend="sec.obsbg.obsconfig"/>.</para>
+   </listitem>
+  </itemizedlist>
+  <para>In all examples, we use:</para>
+  <itemizedlist>
    <listitem>
     <para>
      A user on a local machine (you) who will be called &exampleuser; who
@@ -414,17 +381,8 @@ Cmnd_Alias  OSC_CMD = /usr/bin/osc, /usr/bin/build
     <para>
      An example upstream open source project available at <link
       xlink:href="&gitupstream1;"/>. This project is used as an example and
-     contains source code in the <link
-      xlink:href="https://en.wikipedia.org/wiki/C%2B%2B"
-      >C++ programming language</link>.
+     contains source code in the C++ programming language.
     </para>
-   </listitem>
-   <listitem>
-    <para>A customized system as shown in <xref
-     linkend="sec.obsbg.obsconfig"/>.</para>
-   </listitem>
-   <listitem>
-    <para>An RPM-based operating system like &opensuse; or &sle;.</para>
    </listitem>
   </itemizedlist>
   <para>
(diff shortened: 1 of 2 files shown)


xml/art-obs-beginners-guide.xml
```diff
@@ -618,24 +621,32 @@ Cmnd_Alias  OSC_CMD = /usr/bin/osc, /usr/bin/build
...
+     on distributions like &opensuse;, create a <firstterm baseform="Spec
+     File">spec file</firstterm>. The basic structure should look like this:
     </para>
     <remark>toms 2017-08-09: Should the spec file be referenced by a GitHub
      link? How much detail do we need here?</remark>
     <screen>#
 # spec file for package &gitproject;
 #
-# -- Copyright comitted --
+# -- Copyright committed --
```
tomschr: I guess, this should be omitted. 😁

xml/art-obs-beginners-guide.xml
```diff
@@ -1076,11 +1161,14 @@ Patch0:         &gitproject;_main.diff</screen>
...
-      <uri>&obshome1;</uri> project. For example, the reasons to decline a
-      submit request are usually build errors, style issues, or other
-      problems. Usually, the declined submit request contains a message which
-      contains the reason. Corrections has to be done in a new submit request
-      by superseding the old one.
+      <uri>&obshome1;</uri> project.
+     </para>
+     <para>
+      For example, reasons to decline a
+      submit request are build errors, style issues.
```
tomschr: This is an incomplete sentence. I've added ", or other problems."

xml/art-obs-beginners-guide.xml
```diff
@@ -204,46 +200,45 @@ https://www.suse.com/communities/blog/suse-studio-integration/
  </sect1>
 
  <sect1 xml:id="sec.obsbg.req">
-  <title>Requirements</title>
+  <title>Requirements for Working with &osccmd; on the Command Line</title>
```
tomschr: Isn't it a bit too long?

xml/art-obs-beginners-guide.xml
```diff
@@ -118,19 +118,10 @@ https://www.suse.com/communities/blog/suse-studio-integration/
...
-   The &obs; is accessible either through a Web interface or a command
-   line interface. Both access the OBS API which controls the database,
-   <remark>
-    "Controls" seems like the wrong word. Not sure about a better on right
-    now. - sknorr, 2017-08-15
-   </remark>
-   <remark>toms 2017-08-16: maybe "communicates" instead of "controls"?</remark>
+   The &obs; is accessible either through a Web interface or a command line
+   interface (&osccmd;). Both are clients for the &obsa; API which provides
+   access to the database,
```
adrianschroeter: the webui is meanwhile part of the api, it is only one application which appears as webui or as api depending on the path.

xml/art-obs-beginners-guide.xml
```diff
@@ -156,14 +147,15 @@ https://www.suse.com/communities/blog/suse-studio-integration/
...
-   interface etc.) are relatively specific in purpose and are of lesser
+   interface etc.) are relatively specific in purpose and therefore of lesser
    importance for this guide.
-   <remark>toms 2017-08-16: Hmn, I think the picture does not reveal
-   "more information". </remark>
-   For more information on them, see <xref linkend="fig.obsbg.concept"/>.
+  </para>
+  <para>
+   The schematic in <xref linkend="fig.obsbg.concept"/> shows the components
+   in context.
```
adrianschroeter: I think we need to update this figure, will look later ...

xml/art-obs-beginners-guide.xml
```diff
@@ -175,24 +167,28 @@ https://www.suse.com/communities/blog/suse-studio-integration/
...
-    purpose, defines access control, other repositories, and build targets
-    (operating systems and architectures).
+    A single project can contain several packages and usually serves a specific
+    organizational purpose. Generic access control, related repositories, and
+    build targets (operating systems and architectures) are all defined at
+    the project level.
    </para>
    <para>
     Projects can also contain other projects (subprojects) to structure work.
     Each project name is separated by colons.
```
adrianschroeter: I would not say "contain" here, projects can have subprojects which are complete isolate from it's parent, execpt that parent project maintainers can make themself also maintainer (and therefore gain write access) to the sub project

xml/art-obs-beginners-guide.xml
```diff
@@ -175,24 +167,28 @@ https://www.suse.com/communities/blog/suse-studio-integration/
...
+    <uri>M17N:fonts</uri> which is a subproject of <uri>M17N</uri>.
+    Packages for the Python programming
+    language are available in the <uri>devel:languages:python</uri> project
+    which is a subproject of <uri>devel:languages</uri> which itself is a
+    subproject of <uri>devel</uri>.
    </para>
    <para>
-    As a user, you will build your packages normally in your
+    As a user, you will normally build packages in your
     <firstterm baseform="Home Project">home project</firstterm>, available
```
adrianschroeter: maybe say here "home project area" to include the sub projects? a standard branch is creating a sub project and does not work in the home project directly ... or is this splitting hairs?

xml/art-obs-beginners-guide.xml
```diff
@@ -175,24 +167,28 @@ https://www.suse.com/communities/blog/suse-studio-integration/
...
     in &obsa; as <uri>home:<replaceable>USERNAME</replaceable></uri>.
-    The home project is your working area in &obsa; for defining build targets,
-    uploading and building packages, and installing the results with Zypper
-    or &yast;.
+    Home projects serve as a personal working area in &obsa; to define build
+    targets, upload, build, and download packages.
```
adrianschroeter: ... and allows to create sub projects what happens when temporary working on other people packages.

xml/art-obs-beginners-guide.xml
```diff
@@ -558,26 +555,32 @@ Cmnd_Alias  OSC_CMD = /usr/bin/osc, /usr/bin/build
...
    To create a package from the upstream project, do the following:
   </para>
   <procedure xml:id="pro.obsbg.uc.basicprj">
    <step>
+    <remark>
+     Again, this is very specific to the openSUSE Build Service! Also, iirc
+     refusal to build is not a thing. OBS itself is totally agnostic to that,
+     but when you submit something to openSUSE (so again openSUSE BS-specific),
+     the license will be checked by some totally optional extra web service.
+     - sknorr, 2017-08-16
```
adrianschroeter: right, I don't mind to have something openSUSE specific, but it would be nice if we can mark this as openSUSE specific. Also using the opensuse profile would be good to allow others to generate books without it.

xml/art-obs-beginners-guide.xml
```diff
@@ -618,24 +621,32 @@ Cmnd_Alias  OSC_CMD = /usr/bin/osc, /usr/bin/build
...
+     Create the &packspec;. This file contains metadata and build
+     instructions.
+    </para>
+    <para>
+     As in this example we are building an RPM which can later be installed
+     on distributions like &opensuse;, create a <firstterm baseform="Spec
+     File">spec file</firstterm>. The basic structure should look like this:
     </para>
     <remark>toms 2017-08-09: Should the spec file be referenced by a GitHub
      link? How much detail do we need here?</remark>
```
adrianschroeter: we don't need much details here, a hint where spec files are documented would be nice tough.

xml/art-obs-beginners-guide.xml
```diff
@@ -930,68 +965,83 @@ Fri Aug 23 12:31:41 UTC 2017 - &exampleuser_mail;</screen>
...
+  </remark>
  </sect1>
 
  <sect1 xml:id="sec.obsbg.uc.branchprj">
-  <title>Branching a Project</title>
+  <title>Branching a Package</title>
+  <remark>
+   Is the OBS definition of branch really so narrow? (i.e. just for home
+   project copies?) - sknorr, 2017-08-16
+  </remark>
```
adrianschroeter: not sure if I understand the question right, but you can branch any package in OBS into any project (if you have write permission). branching below home namespace is just the default when no target is specified. These branches are also marked as temporary so there will happen an autocleanup via a delete request later. Or they get deleted when a submission to another project got accepted by default.

xml/art-obs-beginners-guide.xml
```diff
@@ -930,68 +965,83 @@ Fri Aug 23 12:31:41 UTC 2017 - &exampleuser_mail;</screen>
...
-     Apply changes in your branch to try out a different path of development.
+     Applying changes in a branch to try out a different path of development
     </para>
    </listitem>
    <listitem>
     <para>
-     Modify the source code, build it, try it out, and submit back your
-     changes to the original project.
+     Modifying the source code, building it, trying the effect of the
+     changes, and submitting back changes to the original project.
```
adrianschroeter: I would also point out here that this is the case when you don't have write access there. This is the way how you can contribute to any package you see in OBS.

Maybe this is is the most important point so it should be first?

xml/art-obs-beginners-guide.xml
```diff
@@ -1006,30 +1056,56 @@ Patch0:         &gitproject;_main.diff</screen>
     <para>Create the branch from &exampleuser_plain;'s home project:</para>
     <screen>&prompt.user2;<command>osc</command> bco &obshome1; &gitproject;</screen>
     <para>This will create a branched package in &obsa; at <uri>&obsbranch2;/&gitproject;</uri>
-     and will checkout a directory <filename>&obsworkdir2;/branches/home/&obsuser1;/&gitproject;</filename>.</para>
+     and will check out a directory
+     <filename>&obsworkdir2;/branches/home/&obsuser1;/&gitproject;</filename>.
```
adrianschroeter: hm, the path is without collons ... this is not our default ...

xml/art-obs-beginners-guide.xml
```diff
@@ -1006,30 +1056,56 @@ Patch0:         &gitproject;_main.diff</screen>
...
+    <para>Make changes as shown in <xref
      linkend="sec.obsbg.uc.patching"/>.
     </para>
    </step>
    <step>
-    <para>Build your package and check the result:</para>
+    <para>Build the package and check the result:</para>
+    <remark>
+     Why not just build for the default OS with all the default options, i.e.
+     just "osc build"? That would be way more generic. - sknorr, 2017-08-16
```
adrianschroeter: good point, should be mentioned at least as well...

xml/art-obs-beginners-guide.xml
```diff
@@ -1038,26 +1114,34 @@ Patch0:         &gitproject;_main.diff</screen>
     osc submitreq [OPTIONS] DESTPRJ [DESTPKG]
     osc submitreq [OPTIONS] SOURCEPRJ SOURCEPKG DESTPRJ [DESTPKG]
     -->
-    <screen>&prompt.user2;<command>osc</command> submitpac &obshome1;</screen>
+    <screen>&prompt.user2;<command>osc</command> submitpac</screen>
     <para>
-     &obsa; knows where the original project is located, so you do not have
-     to add the project name here.
+     &obsa; usually knows where the original project is located, so you do
```
adrianschroeter: Please use "osc submitreq" here .... the "submitpac" is actually more or less obsolete meanwhile...

You could also write it that way that "osc submitreq is by default submitting back to the location where you branched from. But it is also possible to specify a different target manualy

xml/art-obs-beginners-guide.xml
```diff
@@ -1076,11 +1161,14 @@ Patch0:         &gitproject;_main.diff</screen>
...
-      contains the reason. Corrections has to be done in a new submit request
-      by superseding the old one.
+      <uri>&obshome1;</uri> project.
+     </para>
+     <para>
+      For example, reasons to decline a
+      submit request are build errors, style issues.
+      When declining submit request, the reviewer will usually give a
+      reason. &obsuser2; can then correct their submission using a new submit
+      request that supersedes the previous one.
```
adrianschroeter: or disagreeing with the reason and "osc reopen" the request. Or accepting the refusal by "osc revoke".

Please point out that OBS will show declined requests still as open requests, but for the submitter. That way they shouldn't get lost and stay on the users task page. The final resolution can be revoke, superseded or accepted.

adrianschroeter: maybe point out also that requests usually submit a concrete version of the source, later changes don't get fetched by default. This is unlike github pull requests where you can modify also afterwards with the risk the a required review step might get missed.

xml/art-obs-beginners-guide.xml
```diff
@@ -1130,23 +1218,22 @@ Patch0:         &gitproject;_main.diff</screen>
...
+   &obsa; provides a place which contains all the distribution-specific and
+   architecture-specific versions
+   of successfully built packages. When you create a package
    in your &obsa; home project, all successfully built packages will appear
    under the <uri>&obsdnlurlhome1;</uri> URL.
   </para>
   <para>
+   However, this only true for the home project itself and manually created
+   subprojects but not for subprojects created as a result of branching a
+   package. Branched projects are not published by default.
```
adrianschroeter: correct, but could you point out that developers can still manually download the build results (osc getbinaries). They just don't get a repository that way.

adrianschroeter:
Added quite some more nitpicking comments, but I have to say that I have enjoyed to read it :)

tomschr:
This is great, thanks @adrianschroeter  and @sknorr for reviewing it! 👍 
I will integrate your suggestions today.

xml/art-obs-beginners-guide.xml
tomschr: Ok, in this case, I would like to remove this sentence ("Both are...") as it seems an implementation detail.

tomschr: Hmn, ok, what about this:

```
Both "talk" to the &obsa; API which provides ...
```

xml/art-obs-beginners-guide.xml
tomschr: If you have a more up-to-date image, please send it to me (or replace it yourself, it's an SVG in `images/src/svg/obs-concept.svg`).

xml/art-obs-beginners-guide.xml
tomschr: Ok, do we need to mention this exception about maintainers? 
I would try to avoid to explain another term and would leave this detail to the User Guide.

What about this?

```
Projects can also have other projects (subprojects) to structure work.
They are isolated from their parents projects and can be configured independently.
[...]
```

I'm not sure about the rest ("... can be configured independently.")

xml/art-obs-beginners-guide.xml
tomschr: Ha, now we come to terminology! 😉

I think, that's a matter of definition what a "home project" is: is it just the direct root of the home project, or does it include other subproject as well?

The definition in the glossary is broad enough to include also subprojects:

**Home Project**
Working area in OBS for uploading and building packages. Each home project starts with `home:USERNAME`.

IMHO I would prefer to **not** introduce another term ("home project area") but use the existing one and -- if needed -- add further explanations ("including your branches and/or subprojects etc.").

xml/art-obs-beginners-guide.xml
tomschr: Good point! 👍

What about:

```
Home projects serve as a personal working area in &obsa; to define build
targets, upload, build, and download packages, and to create temporary
subprojects ("branches") to work on other people's packages. 
```

xml/art-obs-beginners-guide.xml
tomschr: It is possible to create two different variants of the same guide ("profiling" is the term we use here).
We can mark some paragraphs, steps, etc. as "openSUSE only" and others as "business" related. With two different DC files, we can create both variants if this is wanted.

xml/art-obs-beginners-guide.xml
adrianschroeter: well, maybe just write "OBS server provides a web interface and an api. The later can be used via osc..." (or alike)

so you avoid the problem if the technical implementation. Does it sound okay to you?

xml/art-obs-beginners-guide.xml
adrianschroeter: I would at least mention that maintainers from upper level can gain write access to subprojects. This  is in the end the entire point of the sub project structure and also a security matter if you are not aware of that.

xml/art-obs-beginners-guide.xml
tomschr: I think you miss the point here. 😉
This is not about how spec files are written in general. In this section, I describe a scenario with a specific spec file. That specific spec file is located on a GitHub project under an openSUSE domain (the "upstream project"). This makes it a bit more easier for the user to follow the details.

So the question is related to this *upstream project*: insert this relationship into the doc as a link or better describe it directly.

xml/art-obs-beginners-guide.xml
adrianschroeter: hm, long sentence .... maybe stop behind the "download packages" and continue with "Users are also permitted further sub projects for isolated work area". 

The branches are just one example where this is used in the end.

xml/art-obs-beginners-guide.xml
adrianschroeter: hm,  I see your point avoiding a new term...

However THE "Home Project" is at least a fixed link in the webui which goes to home:$user directly. So it is at least a fixed term here for this single one.

Maybe we could just speak in plural "the users home projects" to include also the subprojects?

xml/art-obs-beginners-guide.xml
adrianschroeter: right, we use the "opensuse"  and "novell" profile already in our default DC files.

xml/art-obs-beginners-guide.xml
tomschr: Ok, I've rephrase it as follows:

```
The &obs; server provides a Web interface and
an API. The API is used by the &osccmd; command line tool.
```

xml/art-obs-beginners-guide.xml
tomschr: That mean, we have to introduce "maintainers" as a new term. This is one of the terms which I was unsure how to define it (still left in the glossary as "to be defined" 😉).
Do you have a suggestion?

If you think this is important and maintainers has to be mentioned, then I would like to add something about maintainers in the "Projects and Packages" section.

adrianschroeter: not sure if the term "maintainers" is important here, but in the end the people with write access on upper level.

xml/art-obs-beginners-guide.xml
tomschr: > Maybe we could just speak in plural "the users home projects" to include also the subprojects?

I would avoid the plural for home project. IMHO, I fear, that would confuse people, because there is basically only one home project. 😉

For your specific wording, I would prefer "the user's home project and his subprojects". Maybe we need to make clear that subprojects includes also branches (which is just a "special" subproject).

xml/art-obs-beginners-guide.xml
tomschr: Right! Your sentence is a bit strange, so I rewrote it and try to insert the two use cases here (create a subproject to structure work or use a subproject (=branch) to work on another package):

```
Home projects serve as a personal working area in &obsa; to define build
targets, upload, build, and download packages.
Users are also permitted to create subprojects to isolate work from the
parent project or to work on branches from another package.
```

xml/art-obs-beginners-guide.xml
sknorr: Right, i forgot inserting an "or". But if there is a "for example" at the beginning, don't add "other problems" at the end.

xml/art-obs-beginners-guide.xml
sknorr: @adrianschroeter question answered! my question referred to toms's definition of "branch" in the paragraph below.

xml/art-obs-beginners-guide.xml
sknorr: In which case, just a single "m" in the word is fine ^^

xml/art-obs-beginners-guide.xml
sknorr: I know but it is more correct nonetheless. Especially, since there is also a Prerequisites section and those two looked kinda similar from their titles.

xml/art-obs-beginners-guide.xml
tomschr: Ahh, I see! Fixed it, add "or" and remove the "other problems".

xml/art-obs-beginners-guide.xml
tomschr: I know. 😄 

The slashes (instead of the colons) are used on purpose and are explained in "Section 4 Configuring Your System for OBS" (basically setting `checkout_no_colon=1`). IMHO it think this layout is be a bit easier to follow for users and they don't mix it up (checking out a directory that contains colons?).

xml/art-obs-beginners-guide.xml
tomschr: Right. I've moved it as first item and added:

```
Usually, you use this workflow when you do not have write permissions of a project.
```

xml/art-obs-beginners-guide.xml
tomschr: Ok, after discussing this topic with Stefan, I will use the default (with colons).

xml/art-obs-beginners-guide.xml
tomschr: Ok, I've added this distinction, but a bit before (search for the option `--local-package`).

xml/art-obs-beginners-guide.xml
tomschr: Ok, replaced `submitpac` with `submitreq`.

tomschr: Added additional sentence to clarify what happens when you use "osc submitreq" only:

```
Usually, &obsa; you do not have to add the project or package names here.
By default, <command>submitreq</command> without any options submits back
to the package where you branched from.
```

xml/art-obs-beginners-guide.xml
```diff
@@ -1038,26 +1114,34 @@ Patch0:         &gitproject;_main.diff</screen>
...
     </para>
+    <screen>&prompt.user2;<command>osc</command> submitpac &obsbranch2;:&obshome1; &gitproject; &obshome1;</screen>
+    <note>
+     <title><command>osc submitrequest</command> and <command>osc submitpac</command></title>
+     <para>
+      With the <command>osc submitpac</command> subcommand, the branch will
+      be automatically deleted when the submit request has been accepted.
+      To avoid this, use the <command>osc submitrequest</command>
+      subcommand, or add the option <option>--no-cleanup</option> to
+      <command>osc submitpac</command>.
```
tomschr: Right, removed `submitpac` and replaced it with `submitreq`.

xml/art-obs-beginners-guide.xml
tomschr: Ok, added some more sentences to make it more clear:

```
User &obsuser2; can then correct their submission using a new submit
request that supersedes the previous one, disagreeing and reopening
the request, or accepting the refusal and revoke the request.
```

tomschr: > maybe point out also that requests **usually** submit a concrete version of the source, later changes don't get fetched by default.

**Usually**? So this is not always the case?

xml/art-obs-beginners-guide.xml
tomschr: Ok, appended the following sentence:

```
However, you can still download the build results manually with
<command>osc getbinaries</command>. 
```

tomschr:
I think, I've covered all the issues above. The lastest changes can be viewed on docserv, I've triggered a rebuild.

Thanks to all for your very valuable comments! 👍 👍

PR #77, opened 2017-08-30
Redefine package

tomschr:
IMHO, the package definition should contain a vage idea and not to delegate it to other terms. I would propose:

* Use of "General term"
* Remove itemizedlist and use glossseealso

Maybe "preassembled" is not the best term, but I hope you get the idea. 😉


tomschr:
Closing this PR and using #79 instead.

PR #78, opened 2017-08-30
Glossary: rephrase & shorten container definition

tomschr:
IMHO, a definition should be brief. Details should be moved in special, dedicated sections.

Therefore, I've shortened the definition and try to add the most important points. Feel free to change/amend/reject it. 😉

PR #79, opened 2017-08-30
Shorten glossary entries for package/container/image

sknorr:
After feedback from toms, shorten several glossary entries:
* remove information that went beyond a typical glossary entry
* simplelists instead of space-consuming itemizedlists
* still try to keep relations between entries intact
* fix a few oversights from yesterday

@tomschr I'd like to see this one go in instead of #77 & #78

### DIFF
--- a/xml/obs_glossary.xml
+++ b/xml/obs_glossary.xml
@@ -72,30 +72,28 @@
  </glossentry>
  <!-- ############## B ############## -->
  <glossentry xml:id="obs.glos.project_binary">
-  <glossterm>Binary Package</glossterm>
+  <glossterm>Binary Package (Binary)</glossterm>
   <glossdef>
    <para>
-    A binary package (in the &obs; context often called a binary) is
-    an archive file that contains an installable version of software and
-    metadata. Binary packages can also include, for example, configuration
-    files and documentation.
+    An archive file that contains an installable version of software and
+    metadata. The metadata includes references to the dependencies of the
+    main software. Dependencies are packaged as additional binary packages.
    </para>
    <para>
-    Binary packages are one possible result of building on &obsa;.
-    &obsa; supports building multiple types of binary package. Examples
-    include &rpmf; packages and &debf; packages.
-   </para>
-   <para>
-    Alternatively, &obsa; can also build operating system images and
-    containers.
+    Formats of binary packages include &rpmf; and &debf;.
+    In the &obsa; context, binary packages are sometimes also called a
+    <emphasis>binaries</emphasis>.
    </para>
    <remark>
     "Binary package" poses a bit of a question: It is a bit unclear when a
-    package can be considered binary: 1) all archive formats are pretty
-    much always binary-based; 2) interpreted languages actually aren't
-    binary. The term "installable package" comes to mind, however, but it
-    is definitely not in common usage yet, unlike binary package.
-    - sknorr, 2017-08-23
+    package can be considered binary: 1) all archive formats are pretty much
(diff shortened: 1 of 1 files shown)


xml/obs_glossary.xml
```diff
@@ -72,30 +72,32 @@
...
     metadata. Binary packages can also include, for example, configuration
     files and documentation.
    </para>
    <para>
-    Binary packages are one possible result of building on &obsa;.
-    &obsa; supports building multiple types of binary package. Examples
-    include &rpmf; packages and &debf; packages.
+    Unlike containers, binary packages do not include dependencies of the
+    main software. Instead, dependencies are packaged as additional binary
+    packages. Formats of binary packages include &rpmf; and &debf;.
```
tomschr: I think this sentence could be a bit misleading. Actually, there are some *meta data/pointer/references* to dependencies, but not the referenced dependencies itself.

To make this distinction more prominent, I would suggest something like this:

```
Unlike containers, binary packages do only contain meta data about the dependencies, but do not
include them [directly]. Instead, dependencies are packaged as additional binary packages. [...]
```

xml/obs_glossary.xml
```diff
@@ -72,30 +72,32 @@
...
-    - sknorr, 2017-08-23
+    package can be considered binary: 1) all archive formats are pretty much
+    always binary-based; 2) interpreted languages actually aren't binary. The
+    term "installable package" comes to mind, however, but it is definitely
+    not in common usage yet, unlike binary package.
+    Additionally, the usage of "binary" to mean "binary package" in OBS docs
+    is problematic. Everywhere else, a binary is defined as a single
+    executable file (though whether that file is binary or actually
+    human-readable like a script, usually does not matter). - sknorr,
+    2017-08-30
```
tomschr: I would propose to complete avoid the term "binary package" altogether. This just leads to problems. Maybe we could mention this inside a parenthesis as an additional term, but we should "ban" this term for the reason you've stated above.

xml/obs_glossary.xml
```diff
@@ -250,25 +252,17 @@
...
+    An image file that contains a deployable version of software and
+    metadata. Dependencies of the main software are also included, such as
+    additional libraries.
    </para>
    <para>
-    Containers cannot be installed. Instead, they are copied as-is onto the
-    target system (<emphasis>deployed</emphasis>). They can then usually be
-    run directly in that state.
+    Unlike operating system images, containers do not include an operating
+    system. Unlike binary packages, containers are deployed and not installed.
```
tomschr: Maybe avoid the "binary packages" term altogether?

xml/obs_glossary.xml
```diff
@@ -250,25 +252,17 @@
...
    </para>
    <para>
-    You can build containers on &obsa;. &obsa; supports
-    building multiple types of container. Container formats include
-    &appimg;, &docker;, and &snap;. <!-- &flatpak; ... later. -->
-   </para>
-   <para>
-    Alternatively, &obsa; can also build binary packages and operating system
-    images.
+    You can build containers on &obsa;.
```
tomschr: This is interesting and I never thought about to include it. 😉

However, I'm not sure if we should start to add supported features into a definition. Shouldn't be a definition something "pure" and independent of any features?

xml/obs_glossary.xml
tomschr: About the term "Bugowner" (couldn't pinpoint it to the correct location).

Inside the definition there is an xref to a section in the Admin Guide. I think, we should avoid back linking to chapters, sections etc.

sknorr:
merged on cmdln.

PR #84, opened 2017-09-05
Integrate how to build AppImages

tomschr:
I've integrated a section about AppImages. It contains still some remarks which I couldn't solve yet, but I guess, it can be reviewed.

If you could have a look, it would be very helpful. Thanks!

### DIFF
--- a/xml/obs_package_formats.xml
+++ b/xml/obs_package_formats.xml
@@ -12,7 +12,7 @@
  <remark>toms 2017-08-18: list advantages/disadvantages/benefits for each format</remark>
  <remark>toms 2017-08-18: comparison of these formats?</remark>
 
- <sect1>
+ <sect1 xml:id="sec.obs.pkgformats.about">
   <title>About Formats</title>
   <remark>toms 2017-08-18: classification between distribution-dependand and
    independant</remark>
@@ -58,6 +58,9 @@
    <listitem>
     <para>SimpleImage format</para>
    </listitem>
+   <listitem>
+    <para>&obs.ai;</para>
+   </listitem>
   </itemizedlist>
 
   <para> If no build recipe format and binary format are specified in the project
@@ -69,7 +72,7 @@
    adding a <literal>Type</literal> line to the project configuration. </para>
  </sect1>
 
- <sect1>
+ <sect1 xml:id="sec.obs.pkgformats.spec2rpm">
   <title>RPM: Spec</title>
   <para>
    RPM (RPM Package Manager) is used on &opensuse;, &sle;, RedHat, Fedora,
@@ -99,7 +102,7 @@
   </variablelist>
  </sect1>
 
- <sect1>
+ <sect1 xml:id="sec.obs.pkgformats.dsc">
   <title>Debian: Dsc</title>
   <para>
    DEB packages are used on all Debian or Ubuntu based distributions.
@@ -133,7 +136,7 @@
   </variablelist>
  </sect1>
 
- <sect1>
+ <sect1 xml:id="sec.obs.pkgformats.pkg">
   <title>Arch: pkg</title>
   <para>
    &obs.arch; files is used on Arch Linux and its derivates. For building
@@ -158,22 +161,14 @@
   </variablelist>
  </sect1>
 
- <sect1 condition="tbd">
+ <sect1 xml:id="sec.obs.pkgformats.snap" condition="tbd">
   <title>Snap</title>
   <para/>
   <remark>TBD</remark>
   <remark>toms 2017-08-18: see obs_build_containers.xml</remark>
  </sect1>
 
-<!--
- <sect1>
-  <title>Livebuild</title>
-  <para/>
-  <remark>TBD</remark>
- </sect1>
--->
-
- <sect1 xml:id="_kiwi_appliance">
+ <sect1 xml:id="sec.obs.pkgformats.kiwi">
   <title>&kiwi; Appliance</title>
   <para>
    <link xlink:href="https://suse.github.io/kiwi/">KIWI</link> is an OS
@@ -224,7 +219,7 @@
   </para>
  </sect1>
 
- <sect1>
+ <sect1 xml:id="sec.obs.pkgformats.simpleimage">
   <title>&obs.si;</title>
   <para>
    <remark>toms 2017-08-18: FIXME This should be explained better</remark>
@@ -278,7 +273,7 @@ systemctl enable sshd</screen>
   </example>
  </sect1>
 
- <sect1>
+ <sect1 xml:id="sec.obs.pkgformats.appimage">
   <title>&obs.ai;</title>
   <remark>toms 2017-08-18: TODO What is it and what is needed</remark>
  </sect1>
--- a/xml/obs_package_formats.xml
+++ b/xml/obs_package_formats.xml
@@ -276,5 +276,42 @@ systemctl enable sshd</screen>
  <sect1 xml:id="sec.obs.pkgformats.appimage">
   <title>&obs.ai;</title>
   <remark>toms 2017-08-18: TODO What is it and what is needed</remark>
+  <para>
+   An &obs.ai; is a bundled application including libraries that
+   allows to run on a variety of different distributions without
+   further installation or modification. The &obs.ai; contains all
+   libraries that are not already part of the base system.
+   &obs.ai; keeps the underlying operating system untouched, which means,
+   you do not need root permission nor installation.
+
+  </para>
+  <para> For building an image in &obs.ai; you need:</para>
+  <variablelist>
+   <varlistentry>
+    <term><filename>PACKAGE</filename></term>
+    <listitem>
+     <para>
+      The original, untouched source archive.
+     </para>
+    </listitem>
+   </varlistentry>
+   <varlistentry>
+    <term><filename><replaceable>PACKAGE</replaceable>.rpm</filename></term>
+    <listitem>
+     <para>
+      The successfully built RPM package.
+     </para>
+    </listitem>
+   </varlistentry>
+   <varlistentry>
+    <term><filename>appimage.yml</filename></term>
+    <listitem>
+     <para>
+      A YAML file which contains a list of dependencies and installation
+      instructions. Dependencies are expressed as RPM package names.
+     </para>
+    </listitem>
+   </varlistentry>
+  </variablelist>
  </sect1>
 </chapter>
--- a/xml/osc_building.xml
+++ b/xml/osc_building.xml
@@ -166,7 +166,6 @@ The buildroot was: /var/tmp/build-root/openSUSE_Tumbleweed-x86_64</screen>
   <remark>TBD</remark>
  </sect1>
 
-
  <sect1 condition="tbd">
   <title>From Kiwi Files (&kiwi;)</title>
    <!--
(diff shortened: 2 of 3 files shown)


xml/obs_package_formats.xml
```diff
@@ -278,8 +273,45 @@ systemctl enable sshd</screen>
   </example>
  </sect1>
 
- <sect1>
+ <sect1 xml:id="sec.obs.pkgformats.appimage">
   <title>&obs.ai;</title>
   <remark>toms 2017-08-18: TODO What is it and what is needed</remark>
+  <para>
+   An &obs.ai; is a bundled application including libraries that
+   allows to run on a variety of different distributions without
```
sknorr: "to run"->"running" *!*

xml/obs_package_formats.xml
```diff
@@ -278,8 +273,45 @@ systemctl enable sshd</screen>
...
+   further installation or modification. The &obs.ai; contains all
+   libraries that are not already part of the base system.
+   &obs.ai; keeps the underlying operating system untouched, which means,
```
sknorr: unnecessarily long sentence: "untouched, which means,"-> "untouched. This means that"

xml/obs_package_formats.xml
```diff
@@ -278,8 +273,45 @@ systemctl enable sshd</screen>
...
+   you do not need root permission nor installation.
```
sknorr: permission->permissions
nor->for
(better: `...do not need <systemitem class="user">root</systemitem> permissions to install an &obs.ai;.`)

xml/obs_package_formats.xml
```diff
@@ -278,8 +273,45 @@ systemctl enable sshd</screen>
...
  </sect1>
 
- <sect1>
+ <sect1 xml:id="sec.obs.pkgformats.appimage">
   <title>&obs.ai;</title>
   <remark>toms 2017-08-18: TODO What is it and what is needed</remark>
+  <para>
+   An &obs.ai; is a bundled application including libraries that
+   allows to run on a variety of different distributions without
+   further installation or modification. The &obs.ai; contains all
```
sknorr: The->An

xml/obs_package_formats.xml
```diff
@@ -278,8 +273,45 @@ systemctl enable sshd</screen>
...
+   libraries that are not already part of the base system.
+   &obs.ai; keeps the underlying operating system untouched, which means,
+   you do not need root permission nor installation.
+
+  </para>
+  <para> For building an image in &obs.ai; you need:</para>
```
sknorr: `To build and &obs.ai; in &obsa;, you need:`

xml/obs_package_formats.xml
```diff
@@ -278,8 +273,45 @@ systemctl enable sshd</screen>
...
+     <para>
+      The successfully built RPM package.
+     </para>
+    </listitem>
+   </varlistentry>
+   <varlistentry>
+    <term><filename>appimage.yml</filename></term>
+    <listitem>
+     <para>
+      A YAML file which contains a list of dependencies and installation
```
sknorr: Some good source that explains YAML syntax here? Maybe the spec for appimage.yml in particular?

xml/obs_package_formats.xml
```diff
@@ -278,8 +273,45 @@ systemctl enable sshd</screen>
...
+   you do not need root permission nor installation.
+
+  </para>
+  <para> For building an image in &obs.ai; you need:</para>
+  <variablelist>
+   <varlistentry>
+    <term><filename>PACKAGE</filename></term>
+    <listitem>
+     <para>
+      The original, untouched source archive.
```
sknorr: remove "untouched" -- yes, that is usually ideal but:
a) the word "original" is already there
b) it is not even necessary for the tar to be untouched/original/... in any way for it to build successfully

xml/obs_package_formats.xml
```diff
@@ -278,8 +273,45 @@ systemctl enable sshd</screen>
...
+     </para>
+    </listitem>
+   </varlistentry>
+   <varlistentry>
+    <term><filename><replaceable>PACKAGE</replaceable>.rpm</filename></term>
+    <listitem>
+     <para>
+      The successfully built RPM package.
```
sknorr: The->A
maybe add "...created from the source archive" at the end of the line.

xml/osc_building.xml
```diff
@@ -196,5 +195,192 @@ The buildroot was: /var/tmp/build-root/openSUSE_Tumbleweed-x86_64</screen>
   <remark>toms 2017-08-18: integrate content from  obs_build_preinstall.xml;
   using a _preinstallimage file</remark>
  </sect1>
- 
+
+ <sect1 xml:id="sec.obs.building.appimage">
+  <title>As &obs.ai;</title>
+  <para>
+   To create an &obs.ai;, you need:
```
sknorr: hm, this seems like it is the same list as obs_package_formats.xml, more or less. can you create a small XML file with just that list in it? otherwise, we have to maintain it twice.

xml/osc_building.xml
```diff
@@ -196,5 +195,192 @@ The buildroot was: /var/tmp/build-root/openSUSE_Tumbleweed-x86_64</screen>
...
+  </para>
+  <itemizedlist>
+   <listitem>
+    <para>
+     an <filename>appimage.yml</filename> file which acts as a build recipe.
+    </para>
+   </listitem>
+   <listitem>
+    <para>
+     everything else you need from building a RPM package (see <xref
```
sknorr: "from building"->"to build"

xml/osc_building.xml
```diff
@@ -196,5 +195,192 @@ The buildroot was: /var/tmp/build-root/openSUSE_Tumbleweed-x86_64</screen>
...
+   </listitem>
+   <listitem>
+    <para>
+     a <filename>_service</filename> file
+     <remark>toms 2017-09-04: is this really needed? Or is it optional?</remark>
+    </para>
+   </listitem>
+   <listitem>
+    <para>
+     some configurations in your &obsa; project, see below.
```
sknorr: remove "some"
configurations->configuration (because this is a mass noun)
"see below"->xref?

xml/osc_building.xml
```diff
@@ -196,5 +195,192 @@ The buildroot was: /var/tmp/build-root/openSUSE_Tumbleweed-x86_64</screen>
...
+   <listitem>
+    <para>directly from source code.</para>
+   </listitem>
+  </itemizedlist>
+  <para>To build a new &obs.ai; from RPM packages, proceed as follows:</para>
+  <procedure>
+   <title>Building &obs.ai; from RPM Packages</title>
+   <step>
+    <para>
+     Start with building RPM packages as described in <xref
```
sknorr: "start with building"->"build" (this is a procedure, step 1 -- this must be the start)

xml/osc_building.xml
```diff
@@ -196,5 +195,192 @@ The buildroot was: /var/tmp/build-root/openSUSE_Tumbleweed-x86_64</screen>
...
+      linkend="sec.obs.building.spec2rpm"/>. Make sure the packages has
```
sknorr: has->have

xml/osc_building.xml
```diff
@@ -196,5 +195,192 @@ The buildroot was: /var/tmp/build-root/openSUSE_Tumbleweed-x86_64</screen>
...
+    <substeps>
+     <step>
+      <para>
+       To open the meta configuration, run:
+      </para>
+      <screen>&prompt.user;<command>osc</command> meta prj -e &obshome1;</screen>
+      <para>
+       <remark>toms 2017-09-05: Maybe this step should be explained in
+       a separate section/procedure?</remark>
+       If you want to use the Web interface of your &obsa; instance to do
```
sknorr: remove "If you want"

xml/osc_building.xml
```diff
@@ -196,5 +195,192 @@ The buildroot was: /var/tmp/build-root/openSUSE_Tumbleweed-x86_64</screen>
...
+    </para>
+    <substeps>
+     <step>
+      <para>
+       To open the meta configuration, run:
+      </para>
+      <screen>&prompt.user;<command>osc</command> meta prj -e &obshome1;</screen>
+      <para>
+       <remark>toms 2017-09-05: Maybe this step should be explained in
+       a separate section/procedure?</remark>
```
sknorr: you're right, it is a bit distracting to have the same thing twice but using a different UI here.

xml/osc_building.xml
```diff
@@ -196,5 +195,192 @@ The buildroot was: /var/tmp/build-root/openSUSE_Tumbleweed-x86_64</screen>
...
+     </step>
+     <step>
+      <para>
+       To enable &obs.ai; support, add the following XML snippet into the editor:
+      </para>
+      <screen language="xml">&lt;repository name="AppImage">
+  &lt;path project="&obshome1;" repository="<replaceable>PROJECT</replaceable>"/>
+  &lt;path project="OBS:AppImage" repository="AppImage"
+  &lt;arch>x86_64&lt;/arch>
+  &lt;!-- &lt;arch>i586&lt;/arch> -->
```
sknorr: why is this XML commented? remove?

xml/osc_building.xml
```diff
@@ -196,5 +195,192 @@ The buildroot was: /var/tmp/build-root/openSUSE_Tumbleweed-x86_64</screen>
...
+       </menuchoice>.
+      </para>
+     </step>
+     <step>
+      <para>
+       To enable &obs.ai; support, add the following XML snippet into the editor:
+      </para>
+      <screen language="xml">&lt;repository name="AppImage">
+  &lt;path project="&obshome1;" repository="<replaceable>PROJECT</replaceable>"/>
+  &lt;path project="OBS:AppImage" repository="AppImage"
```
sknorr: this is oSBS-specific! make the reader aware.

xml/osc_building.xml
```diff
@@ -196,5 +195,192 @@ The buildroot was: /var/tmp/build-root/openSUSE_Tumbleweed-x86_64</screen>
...
+     </step>
+     <step>
+      <para>
+       Adapt the <replaceable>PROJECT</replaceable> placeholder and the
+       architecture.
+      </para>
+     </step>
+     <step>
+      <para>
+       To save the new configuration, use <keycombo>
```
sknorr: remove "new" (wordy)

xml/osc_building.xml
```diff
@@ -196,5 +195,192 @@ The buildroot was: /var/tmp/build-root/openSUSE_Tumbleweed-x86_64</screen>
...
+        <keycap>:</keycap>
+        <keycap>x</keycap>
+       </keycombo> or leave the editor with <keycombo>
+        <keycap>:</keycap>
+        <keycap>q</keycap>
+        <keycap>!</keycap>
```
sknorr: better remove the vi instructions -- osc uses the $EDITOR variable, not everyone sets that to vi.
maybe setting up an appropriate editor should be in the Configuration part?

xml/osc_building.xml
```diff
@@ -196,5 +195,192 @@ The buildroot was: /var/tmp/build-root/openSUSE_Tumbleweed-x86_64</screen>
...
+       </keycombo>.
+      </para>
+     </step>
+    </substeps>
+   </step>
+   <step>
+    <para>
+     Create a new <filename>appimage.yml</filename> with the following
```
sknorr: remove "new" (already implied in "Create")

xml/osc_building.xml
```diff
@@ -196,5 +195,192 @@ The buildroot was: /var/tmp/build-root/openSUSE_Tumbleweed-x86_64</screen>
...
+     contents:
```
sknorr: contents->content

xml/osc_building.xml
```diff
@@ -196,5 +195,192 @@ The buildroot was: /var/tmp/build-root/openSUSE_Tumbleweed-x86_64</screen>
...
+    - <replaceable>MY_RPM_PACKAGE</replaceable>
+
+script: <co xml:id="co.obs.building.ai.script"/>
+  # instructions
+     </screen>
+     <calloutlist>
+      <callout arearefs="co.obs.building.ai.app">
+       <para>
+        <remark>toms 2017-09-05: any restrictions in naming?</remark>
+        Contains the name of &obs.ai; app. You can use any name you want,
```
sknorr: app->application

xml/osc_building.xml
```diff
@@ -196,5 +195,192 @@ The buildroot was: /var/tmp/build-root/openSUSE_Tumbleweed-x86_64</screen>
...
+        but it is recommended to use lowercase names.
```
sknorr: "it is recommended to use"->"we recommend using"

xml/osc_building.xml
```diff
@@ -196,5 +195,192 @@ The buildroot was: /var/tmp/build-root/openSUSE_Tumbleweed-x86_64</screen>
...
+       </para>
+      </callout>
+      <callout arearefs="co.obs.building.ai.build">
+       <para>Defines the build requirements as RPM package names.</para>
```
sknorr: maybe: as->using ?

xml/osc_building.xml
```diff
@@ -196,5 +195,192 @@ The buildroot was: /var/tmp/build-root/openSUSE_Tumbleweed-x86_64</screen>
...
+     directory under der name <filename><replaceable
+      >APPNAME</replaceable>-<replaceable
+       >VERSION</replaceable>-<replaceable
+        >LIBRARY</replaceable>-<replaceable
+         >ARCH</replaceable>.AppImage</filename>.
+    </para>
+   </step>
+  </procedure>
+  <para>
+   If you want to build &obs.ai;s from source, you need to change the
```
sknorr: remove "If you want"
remove "you need to"

xml/osc_building.xml
```diff
@@ -196,5 +195,192 @@ The buildroot was: /var/tmp/build-root/openSUSE_Tumbleweed-x86_64</screen>
...
+   <listitem>
+    <para>
+     everything else you need from building a RPM package (see <xref
+      linkend="sec.obs.building.spec2rpm"/>).
+    </para>
+   </listitem>
+   <listitem>
+    <para>
+     a <filename>_service</filename> file
+     <remark>toms 2017-09-04: is this really needed? Or is it optional?</remark>
```
adrianschroeter: It is needed only when you specify external resources, like a git repository.

xml/osc_building.xml
```diff
@@ -196,5 +195,192 @@ The buildroot was: /var/tmp/build-root/openSUSE_Tumbleweed-x86_64</screen>
...
+   There are two ways to build an &obs.ai;:
+  </para>
+  <itemizedlist>
+   <listitem>
+    <para>From RPM packages, or</para>
+   </listitem>
+   <listitem>
+    <para>directly from source code.</para>
+   </listitem>
+  </itemizedlist>
```
adrianschroeter: You may also combine the two approaches, meaning including files from rpm packages, but also compile additional files.

xml/osc_building.xml
```diff
@@ -196,5 +195,192 @@ The buildroot was: /var/tmp/build-root/openSUSE_Tumbleweed-x86_64</screen>
...
+     <title>&obs.ai; from RPM Files (<filename>appimage.yml</filename>)</title>
+     <screen language="yaml">app: <replaceable>APPNAME</replaceable> <co xml:id="co.obs.building.ai.app"/>
+
+build: <co xml:id="co.obs.building.ai.build"/>
+  packages:
+    - linuxdeployqt
+
+ingredients: <co xml:id="co.obs.building.ai.ingredients"/>
+  packages:
+    - <replaceable>MY_RPM_PACKAGE</replaceable>
```
adrianschroeter: Can you also add commented out example lines how to integrate upstream git sources?
eg

  git:
    - https://github.com/probonopd/QtQuickApp.git

xml/osc_building.xml
```diff
@@ -196,5 +195,192 @@ The buildroot was: /var/tmp/build-root/openSUSE_Tumbleweed-x86_64</screen>
...
+  <example>
+   <title>&obs.ai; from Source File (<filename>appimage.yml</filename>)</title>
+   <screen language="yaml">app: <replaceable>APPNAME</replaceable>
+
+build:
+  packages:
+    - linuxdeployqt
+    - cmake
+  git:
+    - &gitupstream1;.git
```
adrianschroeter: ah ... here you go ...

please note that this works also with svn, bzr, hg

xml/osc_building.xml
```diff
@@ -196,5 +195,192 @@ The buildroot was: /var/tmp/build-root/openSUSE_Tumbleweed-x86_64</screen>
...
+
+script:
+   - cmake -DCMAKE_INSTALL_PREFIX=/usr .
+   - make
+   - make install DESTDIR=/usr
+   # other lines suppressed</screen>
+  </example>
+  <remark>toms 2017-09-05: do we need a _service file?</remark>
```
adrianschroeter: only when using the git upstream integration.

adrianschroeter:
Did I miss the exact specification between "build" packages and "ingredients"?

BuildRequires vs incorporated packages into the image.

It is also worth to note that the main tooling for building appimages is inside of pkg2appimage package by default. This package contains also a list of packages which get uninstalled at the end, because they are considered base packages always provided by the linux OS by default.

xml/obs_package_formats.xml
tomschr: Fixed.

xml/obs_package_formats.xml
tomschr: Fixed

xml/obs_package_formats.xml
tomschr: Fixed.

xml/obs_package_formats.xml
tomschr: Fixed and used your suggestion in parenthesis.

xml/obs_package_formats.xml
tomschr: I thought of that too, but obviously forgot to add this. Good point! 👍 
Probably http://yaml.org/ should be enough...

xml/osc_building.xml
tomschr: I know, it seems these two sections grow organically. Not sure what's the best solution here. On the one hand, it would be nice to have one dedicated chapter/section which just contains the requisites. But on the other hand, everything in one place would be better...
I need to think about it more or maybe we could discuss this in person.

xml/obs_package_formats.xml
tomschr: Fixed

xml/obs_package_formats.xml
tomschr: Right, fixed.

xml/obs_package_formats.xml
tomschr: Ok, fixed.

xml/obs_package_formats.xml
tomschr: Added `<link xlink:href="http://yaml.org">YAML</link>`

xml/osc_building.xml
tomschr: Fixed

xml/osc_building.xml
tomschr: Ok, fixed.

xml/osc_building.xml
tomschr: Fixed.

xml/osc_building.xml
tomschr: Fixed.

xml/osc_building.xml
tomschr: Fixed

xml/osc_building.xml
tomschr: Ok, added `The <uri>OBS:AppImage</uri> is a project specific to &osbs;.`.

xml/osc_building.xml
tomschr: Well, the idea was, the reader has all(?) possible values and it can remove the comment to enable it. That way, the reader doesn't have to search for the correct value (was it i386, i486, or i586?).

xml/osc_building.xml
tomschr: Fixed.

xml/osc_building.xml
tomschr: That's true, but isn't it the default?

> maybe setting up an appropriate editor should be in the Configuration part?

Probably yes.

xml/osc_building.xml
tomschr: Right, fixed.

xml/osc_building.xml
tomschr: Fixed

xml/osc_building.xml
tomschr: Fixed

xml/osc_building.xml
tomschr: Fixed.

xml/osc_building.xml
tomschr: Right, fixed.

xml/osc_building.xml
tomschr: Fixed.

xml/osc_building.xml
tomschr: Ok, amended the list with another item: `a combination of the previous two methods.`

xml/osc_building.xml
tomschr: See below. 😉
Yes, the idea was to separate these methods to remove clutter. One example with RPM packages only and one example directly from source code.

xml/osc_building.xml
tomschr: Good to know. In this case, the `git` keyword is then `svn`, `bzr`, and `hg`, right?

xml/osc_building.xml
tomschr: Ok, got it, thanks! I've corrected this in the code.

tomschr:
> Did I miss the exact specification between "build" packages and "ingredients"?

There is the `calloutlist` in line 327 after the listing which explains a bit the difference between "build" and "ingredients". However, this can be improved, I guess.

tomschr:
@sknorr @adrianschroeter Thanks for all comments and remarks! 👍 They are integrated now.

Apart from them, I still don't get some details. Maybe you can answer my questions:

1. What is this package `linuxdeployqt` inside the `build/packages` section? What it is used for? Do I have to use it or can I build an AppImage without this?
1. Does the `linuxdeployqt` package has anything to do with Qt?
1. Why is the command `linuxdeployqt` used twice in the [`appimage.yml` file for osc](https://build.opensuse.org/package/view_file/openSUSE:Tools/osc/appimage.yml?expand=1) (line 29+30)?

I've tried to build an AppImage for our example projects (see https://build.opensuse.org/package/show/home:obsgeeko/my-first-obs-package) but it still fails. 😢 Could you have a look and give me some hints what did I miss? Thanks!

adrianschroeter:
linuxdeployqt is one possible way to build appimage containers provided by appimage upstream.

The qt inside the name is only there because it is using itself the qt libraries, but the tool can also be used for non-Qt appimage containers.

IIRC the double usage was needed as workaround for some bug for some time, but should not be needed anymore.

PR #89, opened 2018-01-24
Add the Cloud Upload info to the Admin Guide

mdeniz:


xml/obs_ag_overview_filesystem.xml
```diff
@@ -2780,6 +2840,21 @@ if (-r $hostconfig) {
...
      <para> warden.log - warden log </para>
     </listitem>
+    <listitem>
+     <para> clouduploadserver.log - cloud upload server log </para>
+    </listitem>
+    <listitem>
+     <para> clouduploadworker.log - cloud upload worker log </para>
+    </listitem>
+   </itemizedlist>
+   <para>The following log files for the upload jobs exist inside the <emphasis>/srv/obs/</emphasis> directory:</para>
```
ChrisBr: it is ```/srv/obs/cloudupload`` or ``/bs/cloudupload/``

xml/obs_ag_installation_and_configuration.xml
```diff
@@ -249,6 +249,19 @@ systemctl start obswarden.service</screen>
      not do this on a system connected to an untrusted network or make sure to
      block the ports with a firewall.</para>
    </warning>
+   <sect3 xml:id="_cloud_upload_setup">
+    <title>Cloud Upload Setup</title>
+    <para>In order to setup the Cloud Upload feature you will need to configure the tools required per each cloud provider.</para>
+    <para>
+      Right now we only support the AWS Amazon Cloud (<link xlink:href="https://aws.amazon.com"/>) as provider. For uploading images in AWS two tools are needed the "aws CLI" and the "ec2uploadimg".
```
ChrisBr: Is it necessary to mention which tools we use in the backend? ->

xml/obs_ag_overview_filesystem.xml
```diff
@@ -647,13 +647,58 @@ auto_cleanup_after_days: 30
          <para>see Reference Guide for more information</para>
         </entry>
        </row>
+       <row>
+        <entry align="left" valign="top">
```
ChrisBr: What is this section for?

xml/obs_ag_overview_filesystem.xml
```diff
@@ -647,13 +647,58 @@ auto_cleanup_after_days: 30
...
+        </entry>
+       </row>
       </tbody>
      </tgroup>
     </table>
     <para>Example feature.yml</para>
     <screen>
-features:
-  image_templates: true
+<![CDATA[
```
ChrisBr: Do we need to include the feature flags here? I guess we will remove the feature flag with the release!

xml/obs_ag_installation_and_configuration.xml
mdeniz: If the admin have to configure those... I guess so

xml/obs_ag_overview_filesystem.xml
mdeniz: The table that shows the features flags?

xml/obs_ag_overview_filesystem.xml
mdeniz: Then whenever it is going to be removed this docu should change. Today and now we have a feature flag that controls it.

xml/obs_ag_overview_filesystem.xml
mdeniz: Ok, will change it

xml/obs_ag_overview_filesystem.xml
ChrisBr: I actually don't see the value of that. It is duplicate work or we might even forget to update. The Unstable branch is not supported. In the official released version there should no feature flag at all.

xml/obs_ag_installation_and_configuration.xml
ChrisBr: The tools get automatically installed by the package so I wouldn't mention it here. I would just mention the config files.

mdeniz: Just mentioning the config files and pointing to examples and a page that explains how to config the tool  is not 99% of the same?

xml/obs_ag_overview_filesystem.xml
mdeniz: So, this docu is for the releases? What happen if anybody installs that and is not working because the flag (the default value is disabled).... Does it make sense to have this info until it gets removed?

ChrisBr: Yes, this documentation is for releases. Releases should not contain feature flags if you ask me.

xml/obs_ag_installation_and_configuration.xml
ChrisBr: I'm just wondering if you state the tools explicitly the Admin might wonder if it is necessary to actually install the tools, no?

mdeniz: Ok, then maybe saying that the admin don't need to install the tools will be ok?

xml/obs_ag_overview_filesystem.xml
mdeniz: Ok, I thought that this feature in particular is something the admin can disable to not show this in the web UI if he don't want to provide it. For me is something optional to an OBS instance, so the flag make sense for me.

ChrisBr: To quote PO: "The feature should be actively promoted by not hiding it" -> so that means we intentionally don't want to hide it! Also it does not harm if it is not hidden, the only thing what would happen is that the user gets a flash message with 'No cloud backend configured, please contact the administrator'

mdeniz: Nice quote and I don't agree with... 
Promoting by disturbing end users with an non configured feature seems a little bit strange to me 😈
If the admin won't use this feature never... he will keep "promoting" the feature for the end users, always! Non sense.

ChrisBr: Well, then you should address this to PO 😄 
Anyway, doesn't change anything because if you ask me, feature flags are something for development and not for administrators. And I would prefer to remove all feature flags before we release a new version.

mdeniz: Ok, @adrianschroeter / @mlschroe what do you think about it?

hennevogel: Feature flags are of course something for administrators. Why do I need them in development?

xml/obs_ag_overview_filesystem.xml
```diff
@@ -2827,6 +2902,10 @@ if (-r $hostconfig) {
 │           ├── srtp
 │           └── wget</screen>
   </sect2>
+  <sect2 xml:id="_cloudupload_directory">
+   <title><filename>cloudupload</filename> Directory</title>
+   <para>Info for cloud upload jobs is stored here, it have a subdir named <emphasis>done</emphasis> for storing the already finished jobs.</para>
```
hennevogel: > it have a subdir named

it **has** a subdir named

xml/obs_ag_overview_filesystem.xml
ChrisBr: https://martinfowler.com/articles/feature-toggles.html

To ship code early and fast without interrupting users. We never hided finished features from users / administrators in the past.

mdeniz: In development for trying it out I guess...

mdeniz: >  We never hided finished features from users / administrators in the past.

By **"We"** you mean the build.opensuse.org admins... but in general it could be helpful for other admins of other instances to not show certain features the won't use.

hennevogel: We can not  hide features from admins as they control feature.yml.
We, as admins of the reference server, have hidden features from (*certain*) users.

xml/obs_ag_installation_and_configuration.xml
mdeniz: Done...

xml/obs_ag_overview_filesystem.xml
ChrisBr: > We can not hide features from admins as they control feature.yml.
We actually can by removing the feature switches from the source code (what we actually did)

> As you see in this section, that Moises extended, we even point admins to feature toggles.
 Write

Which was pointless as we removed the features switches from the source code so changing the feature.yml had no effect.

ChrisBr: So far we used feature switches only to hide new features from users before the official release. After the official release we removed the features checks from the source code. If we want to change that in the future we need to talk about it but so far we only used it for this purpose (which I call as development)

adrianschroeter: I tend to agree with Christian in the way that we have already the /configuration route to enable or hide features. Having a second place makes it not easier.

In any case we need to avoid to ask admins to manually enable features after update which should be there by default. means our built-in defaults must fit.

mdeniz: I think we all agree in that maybe we need to use the feature flags only for trying out things for a certain users, but to release a possible optional feature we need to convert that flag into a configuration entry. I mean by moving the calculation of the "flag" from the feature flag to a thing like CONFIG["flag"] that way other admins can disable features for production.

mdeniz: Nevertheless I will keep this until things change. Maybe worth to have a card in our backlog for doing this. What do you think? @hennevogel @ChrisBr

ChrisBr: Fine for me

ChrisBr:
Wondering if it is worth to mention the Third Party authentication for AWS similar to what we have in the wiki https://github.com/openSUSE/open-build-service/wiki/Cloud-uploader#workflow

xml/obs_ag_overview_filesystem.xml
hennevogel: > After the official release we removed the features checks from the source code. 

Sorry this is simply not true. Look at the history of [config/feature.yml](https://github.com/openSUSE/open-build-service/commits/master/src/api/config/feature.yml). All of those switches are still there in the code.

> I tend to agree with Christian in the way that we have already the /configuration route to enable or hide features.

What feature (in the sense of **features** we talk about here, not difference behavior for certain things) can you configure from inside the app? None.

> In any case we need to avoid to ask admins to manually enable features after update which should be there by default.

I really would do it the other way around. Enabling those far reaching features that we have behind a feature flag should not *self-enable* all of the sudden. They have implications. Image templates have to be there, cloud shit has to be configured etc.  It's up to the admin to make sure these prerequisites are there.

We ensure this BTW currently by marking config/features.yml as `%config(noreplace)` in our [spec file](https://github.com/openSUSE/open-build-service/blob/master/dist/obs-server.spec#L599) and features that are not inside the yml file will be considered off.

> So, to release a possible optional feature we need to convert that flag into a configuration entry.

I'm not sure I agree. Of course you can do that and make everything runtime configurable, from inside the app but then you don't need feature flags at all.

Also not everything is runtime configurable...

mdeniz:
Ok, @ChrisBr  @hennevogel please review.  I added the Third Party auth for AWS and a better explanation of why we need it.

xml/obs_ag_overview_filesystem.xml
mdeniz: Beside this conversation..., do you agree with letting this go as it is right now? and make changes after, probably adding first a card in our sprint backlog.

So, this documentation will remain as it is in this PR and we will change it if we change the way we use the feature flags.

mdeniz:
BTW to make it easier to review. This is a screenshot of the setup part:

imagen

xml/obs_ag_installation_and_configuration.xml
```diff
@@ -249,6 +249,47 @@ systemctl start obswarden.service</screen>
...
+      Right now we only support the AWS Amazon Cloud (<link xlink:href="https://aws.amazon.com"/>) as provider.
+    </para>
+    <sect4 xml:id="_aws_authentication_workflow">
+      <title>How does the AWS's authentication work for uploading images?</title>
+      <para>
+        We are going to use the role based authentication provided by Amazon to enable the OBS instance to upload images to other user's accounts.
+      </para>
+      <para>
+        The users will obtain an external ID (automatically created and unique) and the OBS account ID to create an Identity and Access Management (IAM) role.
+        After the user created the role, he needs to provide the Amazon Resource Name (ARN) of the role to OBS which stores it in Cloud::EC2::Configuration.
```
ChrisBr: Sorry for the late review and it is already merged ...

> which stores it in Cloud::EC2::Configuration

I would not include that as it is developer documentation. For the admin it is not relevant in which model / database table this information get stored. I would rather include that the user can set it on ``/cloud/configuration/ec2`` for instance when you want to include that.

Maybe I would also include a sentence that therefore an uploader account is necessary which gets configured in the next step (see AWS authentication credentials setup)

PR #93, opened 2018-03-06
Update admin guide for OBS cloud upload configuration

bgeuken:
* Mention the packages that has to be installed and services that need to be started.
* Fix path to AWS configuration file
* Drop section about the enceladus script we use, since we don't want or expect users to
  change it's default configuration.

### DIFF
--- a/xml/obs_ag_installation_and_configuration.xml
+++ b/xml/obs_ag_installation_and_configuration.xml
@@ -270,18 +270,36 @@ systemctl start obswarden.service</screen>
       </para>
     </sect4>
     <sect4 xml:id="_aws_authentication_setup">
-      <title>AWS authentication credentials setup</title>
+      <title>AWS authentication – Server setup</title>
       <para>
-      For uploading images in AWS two tools are needed internally the <link xlink:href="https://aws.amazon.com/cli">aws CLI</link> and the <link xlink:href="https://github.com/SUSE/Enceladus/tree/master/ec2utils/ec2uploadimg">ec2uploadimg</link> (those are installed as dependencies).
-      For setting this up you have to configure both tools in the machine where the cloud upload workers will be running.
-    </para>
-    <para>
-      The configurations are stored in <emphasis>/etc/obs/cloudupload</emphasis> (the clouduploader home directory).
-      In this directory there is a <emphasis>.ec2utils.conf</emphasis> file which contains e.g. the helper instance AMIs (see this <link xlink:href="https://github.com/SUSE/Enceladus/blob/master/ec2utils/ec2utils.conf.example">example</link>).
-      In the <emphasis>/.aws/credentials</emphasis> file the AWS credentials of the OBS account are stored (see this <link xlink:href="https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-getting-started.html">documentation</link>).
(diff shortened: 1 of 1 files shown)


xml/obs_ag_installation_and_configuration.xml
```diff
@@ -270,18 +270,37 @@ systemctl start obswarden.service</screen>
...
+            <screen>
+              zypper in obs-cloud-uploader
+              zypper in aws-cli
+            </screen>
+          </listitem>
+          <listitem>
+            <para>Start the cloud upload services</para>
+            <screen>
+              rcobsclouduploadworker start
+              rcobsclouduploadserver start
```
bgeuken: @mlschroe That's correct, right? Both services are needed?

bgeuken:
@ChrisBr Please review

xml/obs_ag_installation_and_configuration.xml
```diff
@@ -270,18 +270,37 @@ systemctl start obswarden.service</screen>
...
+        </orderedlist>
+      </para>
+    </sect4>
+    <sect4 xml:id="_aws_authentication_credential_setup">
+      <title>AWS authentication – Credentials setup</title>
+      <para>
+        For uploading images to AWS, OBS is using the <link xlink:href="https://aws.amazon.com/cli">AWS CLI</link> tool.
+        Before you can start uploading your images, you have to enter the AWS credentials to the
+        <emphasis role="strong">/etc/obs/cloudupload/.aws/credentials</emphasis> configuration file.
+        These credentials will then be used by OBS to upload images to AWS.
```
ChrisBr: Hm, I would write:

> These credentials are used to retrieve temporary credentials to upload images to AWS.

or something similar. Because these credentials are not used to upload, only to retrieve the temporary credentials from the ARN provided by user. More information here (maybe even include the link?):
https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_for-user_externalid.html

I think it is important to clarify otherwise it sounds like there are costs involved for this account.

xml/obs_ag_installation_and_configuration.xml
```diff
@@ -270,18 +270,37 @@ systemctl start obswarden.service</screen>
...
-    </para>
-  </sect4>
-
+        Before you can start uploading images to the Amazon Web Services (AWS) you have to:
+        <orderedlist>
+          <listitem>
+            <para>Install the obs-cloud-uploader package</para>
+            <screen>
+              zypper in obs-cloud-uploader
+              zypper in aws-cli
```
hennevogel: :thought_balloon: Why isn't that a dependency of the obs-cloud-uploader package?

xml/obs_ag_installation_and_configuration.xml
hennevogel: yes both are needed

xml/obs_ag_installation_and_configuration.xml
bgeuken: Ah, right. I already asked @ChrisBr whether he could take care of this (done in https://github.com/openSUSE/open-build-service/pull/4624). Just forgot to scratch it from my notes, So it landed here.

I'll change it

PR #95, opened 2018-03-08
Document backend configuration for OBS cloud upload feature

bgeuken:


bgeuken:
@mlschroe Please check if this is correct. Thanks

PR #102, opened 2018-06-14
Fix grammar issues

athos-ribeiro:


xml/obs_source_service.xml
```diff
@@ -410,7 +410,7 @@
 
  <sect1 xml:id="sec.obs.sserv.remove">
   <title>Removing a Source Service</title>
-  <para> Sometimes it is useful to continued to work on generated files
+  <para> Sometimes it is useful to continue work on generated files
```
eduardoj: Nitpick: maybe "continue working" fits better?

Thanks for reviewing the docs.

Ana06: I think it should be `to continue working` or `to continue to work`. I think `to continue work` is wrong, except if `work` is a noun, but here it is a verb.

athos-ribeiro: I thought it was using as a noun, since there was a identical string on `xml/obs_ag_source_service.xml`. I went ahead and changed both strings.

athos-ribeiro:
I believe you merged it between your comment and my last `push`. The requested fix ended up being merged with both changes, as you pointed out! :)

Ana06:
@athos-ribeiro 

>  I believe you merged it between your comment and my last push. The requested fix ended up being merged with both changes, as you pointed out! :)

Yes! I realised you changed already so I edited my comment and merged! :rofl:

PR #104, opened 2018-06-29
document DISTURL handling

adrianschroeter:


adrianschroeter:
no concerns it seems, so I merge now.

PR #108, opened 2018-08-21
Introduces a development environment

hennevogel:
I've got tired of dealing with daps dependencies on my host system. The
`openbuildservice/obs-docu` image is an automated build on hub.docker.com.


docker-compose.override.yml.example
```diff
@@ -0,0 +1,7 @@
+version: '2.1'
+services:
+  obs-docu:
+    build:
+      args:
+        CONTAINER_USERID: 13042
```
ChrisBr: Why 13042 ? Is it some random number? Why not using a more speaking constant like ``TO_BE_REPLACED``?

hennevogel: yo just a random number...

PR #111, opened 2018-09-14
Staging Project Configuration

adrianschroeter:
Initial description of not-yet-implemented staging project setup and configuration.

The design tries to be similar as maintenance project setup and be flexible enough for different kinds of setups (like only adi or no-adi or everything)

### DIFF
--- a/xml/obs_staging_workflow.xml
+++ b/xml/obs_staging_workflow.xml
@@ -1,0 +1,97 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE chapter
+[
+  <!ENTITY % entities SYSTEM "entity-decl.ent">
+    %entities;
+]>
+<chapter version="5.1" xml:id="cha.obs.stagingworkflow"
+ xmlns="http://docbook.org/ns/docbook"
+ xmlns:xi="http://www.w3.org/2001/XInclude"
+ xmlns:xlink="http://www.w3.org/1999/xlink" >
+ <title>Staging &obsa; Workflow</title>
+ <para>
+   Staging projects can be used to test submissions before they get merged
+   into the target project.
+   Alternatives to staging projects are either <xref linkend="obs.glos.develproject"/>
+   or maintenance <xref linkend="obs.glos.incident"/>.
+   Staging projects are used when the target project is no official maintained
+   project as part of the maintenance workflow, but you still want to group and test
+   submissions from multiple sources together. Or don't trust the setup of the
+   devel projects.
+   The staging projects might be managed by a different group of people then
+   the targeting project. We differe between <xref linkend="obs.glos.release_manager"/>
+   and the <xref linkend="obs.glos.staging_manager" /> here.
+ </para>
+ <remark>TBD</remark>
+
+ <sect1 xml:id="sec.obs.stagingworkflow.staging_kinds">
+  <title>Difference between static and dynamic managed staging projects</title>
+  <remark>
+    static projects are the current "letter projects".
+    dynamic projects are the current "adi:X projects".
+  </remark>
+  <para>
+    Building staging projects may be managed different what is defined by the project kind 
+    of their master project, which is either staging_static or staging_dynamic.
+  </para>
+  <itemizedlist>
+   <listitem>
+    <para>
+     Static staging projects will always be kept after merging or reverting submissions.
+     They are usually used for the base core of larger projects and usually setup in a way
+     that they do a full bootstrap if the submission makes this necessary.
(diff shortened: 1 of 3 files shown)


adrianschroeter:
[Make the staging dashboard configurable](https://trello.com/c/z7ORvgFA/1379-make-the-staging-dashboard-configurable)

adrianschroeter:
An alternatie solution for the two different staging project kinds would be to have just one "staging" kind and allow to enable the project cleanup via an attribute called "OBS:CleanupStagingProjects"

mdeniz:
How can you document something that is not yet implemented? :thinking:
The design can change and then this is a nonsense...

adrianschroeter:
On Montag, 17. September 2018, 11:18:18 CEST wrote Moisés Déniz Alemán:
> How can you document something that is not yet implemented? :thinking:

it is IMHO good common praxis (similar like test driven development) to look
first at the POV of the user.

> The design can change and then this is a nonsense...

sure, we need to adapt it then. But this way we can avoid the situation 
like with the check api where it just clashes when all is implemented.
(and I suppose Saray and Danny are not very experienced with the maintenance
 workflow and setup yet, so I thought it might be helpful)

-- 

Adrian Schroeter
email: adrian@suse.de

SUSE Linux GmbH, GF: Felix Imendörffer, Jane Smithard, Graham Norton, HRB 21284 (AG Nürnberg) 
 
Maxfeldstraße 5                         
90409 Nürnberg 
Germany

xml/obs_staging_workflow.xml
```diff
@@ -0,0 +1,97 @@
...
+   Staging projects can be used to test submissions before they get merged
+   into the target project.
+   Alternatives to staging projects are either <xref linkend="obs.glos.develproject"/>
+   or maintenance <xref linkend="obs.glos.incident"/>.
+   Staging projects are used when the target project is no official maintained
+   project as part of the maintenance workflow, but you still want to group and test
+   submissions from multiple sources together. Or don't trust the setup of the
+   devel projects.
+   The staging projects might be managed by a different group of people then
+   the targeting project. We differe between <xref linkend="obs.glos.release_manager"/>
```
dmarcoux: We differ

xml/obs_staging_workflow.xml
```diff
@@ -0,0 +1,97 @@
...
+ <remark>TBD</remark>
+
+ <sect1 xml:id="sec.obs.stagingworkflow.staging_kinds">
+  <title>Difference between static and dynamic managed staging projects</title>
+  <remark>
+    static projects are the current "letter projects".
+    dynamic projects are the current "adi:X projects".
+  </remark>
+  <para>
+    Building staging projects may be managed different what is defined by the project kind
```
dmarcoux: This sentence is not clear, I don't understand it.

xml/obs_staging_workflow.xml
```diff
@@ -0,0 +1,97 @@
...
+  </para>
+  <itemizedlist>
+   <listitem>
+    <para>
+     Static staging projects will always be kept after merging or reverting submissions.
+     They are usually used for the base core of larger projects and usually setup in a way
+     that they do a full bootstrap if the submission makes this necessary.
+     The cleanup will just remove the package sources when a submit request gets a final
+     state (what can be for example accepted, decline, revoked), but keep the project and
+     it's setup. This allows additional test packages inside like product builds. Also
```
dmarcoux: its setup

xml/obs_staging_workflow.xml
```diff
@@ -0,0 +1,97 @@
...
+   <listitem>
+    <para>
+     Create the staging master project by creating any new project with specified kind "staging_static"
+     or "staging_dynamic" depending on your needs.
+     It is recommended, but not required to create it as subproject.
+    </para>
+   </listitem>
+   <listitem>
+    <para>
+     You can define <xref linkend="obs.glos.staging_manager" /> by assigning these persons or groups
```
dmarcoux: these people or groups.

It could even be users, I think it would fit better.

xml/obs_staging_workflow.xml
```diff
@@ -0,0 +1,97 @@
...
+     with the maintainer role in the staging project. You may not define anyone if it is the same group
+     of persons who manage the target project and the staging master project is a sub project of it.
```
dmarcoux: Same thing here with `persons`. It's either people or users.

dmarcoux:
I think it helps newcomers like me. At the same time, I see this more as a work in progress. It will change depending on what we do in the staging workflow.

hennevogel:
This is now implemented in some other way. Documentation exists, starting with #113 last year

PR #117, opened 2018-11-23
Add deletion of a staging to the best practices

krauselukas:
screenshot-2018-11-23 using the obs web ui best practice guide


xml/obs_best_practice_staging_workflow.xml
```diff
@@ -80,6 +80,23 @@
       </listitem>
     </itemizedlist>
   </sect2>
+  <sect2>
+    <title>Delete a Staging</title>
+    <para>
+      By clicking on the delete button on the Staging index page, we are able
+      to delete a Staging. By selecting the associated staging projects
+      in the appearing window, we are able to delete them as well. If not selected,
```
dmarcoux:
```suggestion
      in the modal window, we are able to delete them as well. If not selected,
```

krauselukas: @dmarcoux i allways thought that "modal" is just a bootstrap term, but it actually makes sense :D i changed it

dmarcoux: It should be `modal window`, not `appearing modal`.

krauselukas: :sweat: done

PR #128, opened 2019-04-05
Edited OBS obs_best_practice_basics.xml to allow for less words and

leapinggreenlizard:
perhaps better clarification


sknorr:
Thanks, but this does not look right. Closing for now, feel free to fix and re-open.

PR #132, opened 2019-11-16
F/movetoadmguide

ghost:
@[](url)


adrianschroeter:
hey, I largely agree with your changes, just

* Dropping Stefan Knorr as author is not so nice
* A generic highlevel OBS server architecture should remain in reference guide, since it is important to know  for understand the interface and work job concepts at least.
* Could you maybe squash your commits a bit? for example the "delete me"?

hennevogel:
@sknorr are you going to work on this?

sknorr:
Yes, I still have it on the list. But not for today.

sknorr:
This is hopelessly bitrotted at this point, so it does not make sense to try to rebase anymore. I have however tried to take at least a few things from this PR.

PR #138, opened 2020-01-27
added docs for running obsworker with kvm backend on k8s

sjamgade:


tomschr:
@sjamgade Thanks for your contribution. :+1: 

You've wrote the content in HTML, however, the documentation is written in DocBook5 format. If you could rewrite the tags to DocBook5 would be great. That makes it possible to merge and integrate it into the existing documentation. Just use one of the many examples in the `xml` folder as a template.

tomschr:
@sjamgade Thanks Sumit for your contribution! :+1:  It validates and builds, so let's ship it!

Issue #139, opened 2020-02-05
Default daps example in README.md doesn't work


wengel:
The default daps example in README.md results in error:

#> daps -d DC-obs-beginners-guide validate

ERROR: Wrong path for the DC-file
       (provided on the command line: DC-obs-beginners-guide)

The file DC-obs-beginners-guide doesn't exist anymore.
Some other files might be used instead, for example DC-obs-user-guide.

I'm creating a pull-request for that.

wengel:
Created [pull-request](https://github.com/openSUSE/obs-docu/pull/140)

tomschr:
Thanks, :+1:  this is now merged (commit 37150b4).

PR #140, opened 2020-02-05
Change example for building books since DC-obs-beginners-guide doesn't exist anymore.

wengel:
Using the daps example as shown in README.md results in [error](https://github.com/openSUSE/obs-docu/issues/139) because the file DC-obs-beginners-guide doesn't exist anymore.

Changing it to DC-obs-user-guide, which exists.


tomschr:
Thanks @wengel for the contribution! Looks good to me! :+1:

PR #143, opened 2020-02-27
Admin Guide: enable services before start

cschneemann:
The text appears to mean 'systemctl enable' the services before starting
them.


tomschr:
Thanks @cschneemann for your contribution! Much appreciated! :+1:

PR #145, opened 2020-05-04
Remove references to non-existent books in README

saraycp:
The books that were merged into User Guide are not mentioned in README
anymore.


tomschr:
Thanks @saraycp you are absolutely right. :+1: Will merge it. Thanks for your contribution!

PR #147, opened 2020-05-04
Update documentation of staging workflow via web UI

saraycp:
Add an extended explanation of the Staging Workflow concept.
Update texts, add some missing sections and update images.


xml/obs_best_practice_staging_workflow.xml
```diff
@@ -7,26 +7,117 @@
...
+    and hardware architectures to build the packages on. Those continuous
+    building processes ensure that the packages are always working for the
+    different setups.
   </para>
   <para>
-    All the Staging workflow starts in the tab 'Staging' which can be found on
-    the project's page. It will take you to the first step to create a new
-    Staging or to the dashboard if the Staging already exists.
+    The maintenance of those packages can be made on a collaborative way via
+    Build Service. As shown in the following graphic, the maintainer can create
```
danidoni, line 22: I think we should use `diagram` here.

danidoni:
Wow, such a great work! :dog: :clap:

xml/obs_best_practice_staging_workflow.xml
saraycp: Thanks, updated @danidoni

PR #150, opened 2020-06-16
add clarification for linked package rebuild and ccache usage

sjamgade:


xml/obs_build_config.xml
```diff
@@ -193,6 +193,9 @@ Keep: debbuild</screen>
            <para>Configure usage of ccache for build of packages mentioned by VALUE. Ccache package will be automatically
                installed and configured. The directory <filename>/.ccache/</filename> will be configured as cache dir.
                To configure ccache, the file <filename>/.ccache/ccache.conf</filename> can be modified as part of the build process by the <envar>$BUILD_USER</envar> environment variable. </para>
+
+           <para>if ccache is enabled but there is no archive present (eg: in case of newly branched packages, wiped binaries etc).
```
dmarcoux, line 197:
```suggestion
           <para>If ccache is enabled, but there is no archive present (e.g.: in case of newly branched packages, wiped binaries etc...).
```

xml/obs_build_config.xml
```diff
@@ -193,6 +193,9 @@ Keep: debbuild</screen>
...
+           The archive of the, same-named package from repository paths configured in project's meta, will be used.</para>
```
dmarcoux, line 198:
```suggestion
           The archive of the homonymous package from repository paths configured in the project's meta will be used.</para>
```

https://www.collinsdictionary.com/dictionary/english/homonymous

xml/obs_build_config.xml
tomschr: I would suggest to replace "e.g." with "for example".

xml/obs_build_config.xml
```diff
@@ -193,6 +193,10 @@ Keep: debbuild</screen>
            <para>Configure usage of ccache for build of packages mentioned by VALUE. Ccache package will be automatically
                installed and configured. The directory <filename>/.ccache/</filename> will be configured as cache dir.
                To configure ccache, the file <filename>/.ccache/ccache.conf</filename> can be modified as part of the build process by the <envar>$BUILD_USER</envar> environment variable. </para>
+
+           <para>If ccache is enabled, but there is no archive present (e.g., in case of newly branched packages, wiped binaries etc...).
```
tomschr, line 197: "e.g." is not always known, so better replace it with "for example":

```suggestion
           <para>If ccache is enabled, but there is no archive present (for example, in case of newly branched packages, wiped binaries etc...).
```

xml/obs_build_config.xml
```diff
@@ -193,6 +193,10 @@ Keep: debbuild</screen>
...
+               The archive of the, same-named package from repository paths configured in project's meta, will be used. 
+           For this archive the architechture will be taken into consideration. And all mentioned paths will be tried in top-to-bottom approach.</para>
```
tomschr, line 199:
```suggestion
           For this archive the architecture will be taken into consideration. All mentioned paths will be tried in top-to-bottom approach.</para>
```

xml/obs_build_config.xml
sjamgade: sorry I force-pushed before looking here for suggestions. 

That is a good one thanks for that, however I am not sure how many others will have to look at dictionary for that word. :-)

xml/obs_build_config.xml
```diff
@@ -193,6 +193,10 @@ Keep: debbuild</screen>
            <para>Configure usage of ccache for build of packages mentioned by VALUE. Ccache package will be automatically
                installed and configured. The directory <filename>/.ccache/</filename> will be configured as cache dir.
                To configure ccache, the file <filename>/.ccache/ccache.conf</filename> can be modified as part of the build process by the <envar>$BUILD_USER</envar> environment variable. </para>
+
+           <para>If ccache is enabled, but there is no archive present (for example, in case of newly branched packages, wiped binaries etc...).
```
tomschr, line 197: Sorry, I still don't get it. I think, this is an incomplete sentence. Usually you have the structure "if bla, then blub.": "if the weather is cold, I will take my sweater with me." However, in this case you have "if bla, but..".

What happens when the ccache is enabled? If you can answer this question, then you can add it to the sentence.

sjamgade: Actually the sentence turned out to be too long so I broke it with a .(fullstop)
Fire me an alternative and I am pretty sure we can include it :-)

sjamgade: If ccache is enabled, but there is no  archive present **from an earlier build**. ..... something on these lines

tomschr: You're missing my point. :-) The sentence is still incomplete. Your sentence should be "if bla, **then** what". However, the sentence stops after the if. You are missing the **then** part.

When you start a sentence with a condition ("if something..."), the reader expects what happens when this condition is true. That's the then clause. For example:

    If there are clouds in the sky,

would be only the if clause. However, that line alone doesn't make any sense. Maybe in poetry, but we are not writing that. :wink: That incomplete line needs a then-clause. One example could be:

    If there are clouds in the sky, then I will take my umbrella with me.

You are adding stuff at the if part, but the sentence needs the then part.
I can't give you an alternative as I don't know what are you trying to tell us. :-)

sjamgade: ```if ccache is enabled, but there is no archive present from an earlier build (for example, in case of newly branched packages, wiped binaries etc...) then the archive of the, same-named package from repository paths configured in project's meta, will be used.```

sjamgade: I the sentence too long ? :thinking:  or where could I break it :-)  And does it help understand now after the `then` part ?

tomschr: Yes, it's a bit long. :-) But that doesn't matter, we can always shorten or rewrite long sentences.

Sorry, but I'm still lost. Maybe we could clarify some things first, before we tackle the sentence length:

* Where does the word "ccache" in the sentence "if ccache is enabled..." refers to?
  To the directory `~/.ccache/`? A directory cannot be enabled, it's just there.
  Or are you referring to the **parameter** `usecache` (some line before inside `<term>`)? In that
  case you should start: "If the parameter usecache is set to true/on/1/whatever, then ..."

* Archive vs. archive name?
  It seems to me, you use the concept of "archive" and "archive names" as synonyms in this sentence, right? For me, an "archive" is just a file sitting somewhere. It has contents and a structure---and a filename.
  In contrast, an "archive name" is basically just a string, a filename of an archive.
  I'm confused what are you referring to.

* Condition(s) (if-else-clauses)
  So if I understood you correctly, you try to write a sentence which looks as (Python) code similar to this:

        if ccache is True and not exists(archive):
            # use name? of... what?
        else:
            # ???
   
   However, the sentences says nothing about when the previous condition is False. This means, the code is executing the else clause.
   Does that make sense to you? Could we add the missing "else" part? One part could be to do that in a list:

    * If ccache is enabled:
        \# list all the results
    * If ccache is disabled:
       \# list all the results

sknorr: This is not all that great, but as a suggestion, this might work:

```
When a package has been newly branched or its binaries have been wiped, no
ccache archive will be available initially.
In such cases, OBS will try to use any available ccache archives of packages
of the same name from the repositories configured in the project's meta
configuration.
```

sknorr:
Hello @sjamgade I just read that you're moving on :sob: -- could you check whether there is anything from your POV that needs to be done in this PR? If not, I'd just try to get it merged soonish.

sknorr:
Since #179 just went in, I am going to close this one.

PR #151, opened 2020-07-02
Fix DoD concept URL

zeroheure:


tomschr:
Thanks @zeroheure for your contribution! :+1:

PR #153, opened 2020-07-15
Cleanup

adrianschroeter:
remove parts pointing to wrong projects with outdated distros and misleading informations...


sknorr:
Thanks! Looked good, merged.

PR #154, opened 2020-07-28
fix some documentation regarding service modes

adrianschroeter:
* enable and fix the table again
* introduce "manual" mode

### DIFF
--- a/xml/obs_source_service.xml
+++ b/xml/obs_source_service.xml
@@ -147,11 +147,6 @@
   <para> Each service can be used in a mode defining when it should run
    and how to use the result. This can be done per package or globally for an
    entire project. </para>
-<!--
-   toms 2017-09-01:
-   This is the table for the variablelist below. It's a good idea, but
-   this table raises more question than it solves.
-   Commenting for the time being.
   <table>
    <title>Service Modes</title>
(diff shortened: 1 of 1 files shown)


xml/obs_source_service.xml
```diff
@@ -284,15 +279,19 @@
...
       the definition as part of the service definition. Or it can be used to
       define the way how to generate the sources and doing so by manually calling
-      <command>osc service runall</command> The result will
-      get committed as standard files again.
+      <command>osc service rundisabled</command>.
+      The result will get committed as standard files again.
+      NOTE: it did only exist as "disabled" before OBS 2.11, but "manual" is the better
+            matching alias name for its usage. The osc client may do have different
+            behaviour in future between manual and disabled.
      </para>
```
sknorr, line 295: This would probably be better as a real note:

```suggestion
     </para>
     <note>
      <title>This Valie Is an Alias for "disabled"</title>
      <para> 
       This option was called "disabled" in earlier versions of OBS.
       In OBS 2.11, it was aliased to "manual" to better reflect usage.
       In the future, the osc client may have different behavior in future
       between "manual" and "disabled".
     </para>
    </note>
```

PR #161, opened 2020-08-19
add docker/skopeo example for disturl inspection

adrianschroeter:


xml/obs_glossary.xml
```diff
@@ -338,6 +338,8 @@
     A DISTURL is structured as <command>obs://OBS_NAME/PROJECT/REPOSITORY/REVISION-PACKAGE</command>.
     It can be shown for example via
     <command>rpm -q --qf '%{DISTURL}\n' kernel-default</command>
+    <command>podman inspect IMAGE_ID | grep org.openbuildservice.disturl</command>
+    <command>skopeo inspect docker://.... | grep org.openbuildservice.disturl</command>
    </para>
```
sknorr, line 343: Hi -- I think it would be useful to have some use-case context for each of the options. Be aware that I don't really know about the use cases for each command, so these use cases are made up--please fix :)

```suggestion
   </para>
   <itemizedlist>
    <listitem>
     <para>
      RPM packages: <command>rpm -q --qf '%{DISTURL}\n' kernel-default</command>
     </para>
    </listitem>
    <listitem>
     <para>
      locally built container images:
      <command>podman inspect IMAGE_ID | grep org.openbuildservice.disturl</command>
     </para>
    </listitem>
    <listitem>
     <para>
      container images built on &obs;:
      <command>skopeo inspect docker://.... | grep org.openbuildservice.disturl</command>
     </para>
    </listitem>
   </itemizedlist>
```

adrianschroeter: The glossar is IMHO not the right place for usecases...

xml/obs_glossary.xml
```diff
@@ -337,10 +337,27 @@
...
+   <itemizedlist>
+    <listitem>
+     <para>
+      RPM packages:
+      <command>rpm -q --qf '%{DISTURL}\n' PACKAGE_NAME</command>
+     </para>
+    </listitem>
+    <listitem>
+     <para>
+      locally built container images:
```
tomschr, line 350: Start with a capital letter:

```suggestion
      Locally built container images:
```

xml/obs_glossary.xml
```diff
@@ -337,10 +337,27 @@
...
+      <command>podman inspect IMAGE_ID | grep org.openbuildservice.disturl</command>
+     </para>
+    </listitem>
+    <listitem>
+     <para>
+      container images built on &obs;:
```
tomschr, line 356: Same here:

```suggestion
      Container images built on &obs;:
```

tomschr (changes requested):
@adrianschroeter Just some minor suggestions. 🙂

xml/obs_glossary.xml
sknorr: In that case, you should probably remove all the commands there, instead of confusing readers with context-free commands.

PR #168, opened 2020-09-16
Add updated SVG version of communication flow image

sknorr:
This pulls out the updated OBS communication flow image from my 2017
branch f/movetoadmguide (which is now hopelessly outdated).

I haven't edited the image since 2017, so I hope it still makes sense--best check via GitHub's rich diff.


adrianschroeter:
I would move the "Download repositories" outside of the frontend and put them into an own "stage server" bubble.

Otherwise I would maybe state that it is simplified (since quite some parts are missing), but looks fine. And we should not over-complicate things here...

adrianschroeter (changes requested):
move repositories into stage server.

sknorr:
Those issues should now be fixed. I have now also added another commit to update the accompanying verbal description, both in terms of language and formatting (it's now a list, much like I did in the old branch).

adrianschroeter:
hm, I do not see it moved in the .png file?

sknorr:
The old PNG is deleted. The new image is an SVG. However, DAPS will create a PNG on the fly when needed (e.g. currently, the HTML version embeds the image as a PNG; whereas PDF gets the unconverted SVG).

adrianschroeter:
Ic.

I would approve it but github is not showing a dialog atm. Please proceed neverthless

sknorr:
Thanks! Merged.

PR #170, opened 2020-09-25
Add further new publish flags

adrianschroeter:


xml/obs_build_config.xml
```diff
@@ -398,6 +398,19 @@ Keep: debbuild</screen>
                 build results get published immediately after the build is finished.</para>
         </listitem>
        </varlistentry>
+       <varlistentry>
+        <term><parameter>nofailedpackages</parameter> (OBS 2.11 or later)</term>
+        <listitem>
+           <para>Block publishing if any build result was failed, broken or unresolvable. Only consistent succeeding builds 
+                 get published. Please note that this gets only evaluated per scheduler architecture.</para>
```
sknorr, line 405: I have a few issues understanding the wording here:

* `consistent succeeding builds`
  * Is most likely a grammar error (it's an adjective but probably needs to be an adverb). If it's meant to be an adjective, you'd need to add a comma: `consistent, succeeding builds`.
  * What is meant by that--`consistently successful builds` or maybe `reproducibly successful builds`?
  * I am not quite sure what happens when the build fails--will older builds of the same package be unpublished? Maybe that should be said more explicitly.

* I am also not sure how to interpret the final sentence, "Please note...". It sounds as if the success is of package build is evaluated for each architecture and each build. If that impression is true, I am not sure why the word "only" is used in that sense--i.e. when is this condition not re-evaluated?

xml/obs_build_config.xml
```diff
@@ -398,6 +398,19 @@ Keep: debbuild</screen>
...
+        </listitem>
+       </varlistentry>
+       <varlistentry>
+        <term><parameter>withreports</parameter> (OBS 2.11 or later)</term>
+        <listitem>
+          <para>Publish also internal content trackings files (.report).</para>
```
sknorr, line 411: Order of `publish` & `also` sounds a bit off, so:

```suggestion
          <para>Also publish internal content tracking files (<literal>.report</report> files).</para>
```

sknorr (commented):
Thanks! Unfortunately, I have a few issues following the description of the first new item. :)

xml/obs_build_config.xml
```diff
@@ -398,6 +398,19 @@ Keep: debbuild</screen>
...
+        <term><parameter>nofailedpackages</parameter> (OBS 2.11 or later)</term>
+        <listitem>
+           <para>Block publishing if any build result was failed, broken or unresolvable. Only consistent succeeding builds 
+                 get published. Please note that this gets only evaluated per scheduler architecture.</para>
+        </listitem>
+       </varlistentry>
+       <varlistentry>
+        <term><parameter>withreports</parameter> (OBS 2.11 or later)</term>
+        <listitem>
+          <para>Also publish internal content tracking files (<literal>.report</report> files).</para>
```
sknorr, line 411: This change introduces a minor validation issue:

```suggestion
          <para>Also publish internal content tracking files (<literal>.report</literal> files).</para>
```

xml/obs_build_config.xml
adrianschroeter: solved by removing confusing sentence & word.

xml/obs_build_config.xml
adrianschroeter: fixed

xml/obs_build_config.xml
sknorr: Iiuc now, the final sentence could be clearer still. Suggestion feature weirdly does not work, but I'd suggest replacing l.404,405 with:

```
           <para>
            Block publishing if any build result was failed, broken or unresolvable.
            This is evaluated individually for each architecture.
            That means, a package can be published for an architecture on which it builds,
            even if it fails to build on another architecture.
           </para>
```

adrianschroeter: changed

PR #175, opened 2020-11-11
remove invalid statement

adrianschroeter:
services mode=disabled exists since the beginning of the time
(of OBS services)


dcermak (approved):
:+1:

PR #190, opened 2021-04-21
Add documentation of the obsgendiff buildflag

dcermak:
~~@adrianschroeter Unfortunately I have no idea what this flag actually does~~

@adrianschroeter I've tried to document the flag to the best of my knowledge.

cc @Vogtinator

### DIFF
--- a/xml/obs_build_config.xml
+++ b/xml/obs_build_config.xml
@@ -233,6 +233,31 @@ Keep: debbuild</screen>
         </para>
        </listitem>
       </varlistentry>
+      <varlistentry>
+        <term><parameter>obsgendiff</parameter></term>
+       <listitem>
+        <para>
+          OBS can run an external program that has access to the current build
+          and the previously successful result, e.g. to generate a difference or
+          a changelog from the diff.
+        </para>
+        <para>
+          OBS will run all scripts in
+          <filename>/usr/lib/build/obsgendiff.d/</filename> on the build host
+          (not in the <emphasis>%buildroot</emphasis>) when this flag is set.
+          If one of the scripts fails to run or no scripts are found, then the
+          overall build fails. I.e. if <literal>BuildFlags: obsgendiff</literal>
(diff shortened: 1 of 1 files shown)


Vogtinator:
All it actually does is: https://github.com/openSUSE/obs-build/blob/e9765d9a64b3c5055701ca9157d426d1ea3610a2/build-recipe#L210

So pretty much all the logic is in `release-compare` and so the documentation here is actually about that (specific) case.

dcermak:
I've reworded the text to make this part clearer.

xml/obs_build_config.xml
```diff
@@ -233,6 +233,32 @@ Keep: debbuild</screen>
         </para>
        </listitem>
       </varlistentry>
+      <varlistentry>
+        <term><parameter>obsgendiff:</parameter></term>
+       <listitem>
+        <para>
+          OBS can run a program to compare the result of the current build with
```
Vogtinator, line 240: I don't think it actually does any comparison, it only extracts the data for external tools to be able to do so

xml/obs_build_config.xml
```diff
@@ -233,6 +233,32 @@ Keep: debbuild</screen>
...
+          the previous result, e.g. to generate a difference or a changelog from
+          the diff.
+        </para>
+        <para>
+          OBS will run all scripts in
+          <filename>/usr/lib/build/obsgendiff.d/</filename> in the buildroot
```
Vogtinator, line 246: Might be confused with the rpm `%buildroot`

xml/obs_build_config.xml
```diff
@@ -233,6 +233,32 @@ Keep: debbuild</screen>
...
+          when this flag is set (which contains the previous build and current
```
Vogtinator, line 247: I don't think it contains the previous build (ignoring the "oldpackages" for `build-compare`)

PR #194, opened 2021-05-18
Review of the Build Recipe section

libnoon:


xml/art-obs-beginners-guide.xml
```diff
@@ -56,45 +56,44 @@ https://www.suse.com/communities/blog/suse-studio-integration/
...
-       <emphasis>installation requirements</emphasis>.
+        You need to indicate how to produce the files that will be
+        part of the package: where to get the sources, and which
+        command(s) to run, e.g. <command>make</command>.
+      </para>
+      <para>
+        The build commands generally requires that some other packages
+        are installed beforehand; for example, to build a C++ program,
+        you need a C++ compiler.  You need to indicate this list of
+        packages: the <emphasis>build requirements</emphasis>.
```
adrianschroeter, line 82: IMHO this is way to broad for our documentation here. We do not want to document package formats in general here, just concepts and OBS specifics.

xml/art-obs-beginners-guide.xml
```diff
@@ -56,45 +56,44 @@ https://www.suse.com/communities/blog/suse-studio-integration/
...
      </formalpara>
-     <para>
-      Build requirements are dependencies which are needed during the
-      build process in &obsa;. For example, a C++ program needs
-      a C++ compiler.
-     </para>
-     <para>
-      Installation requirements are dependencies which are needed when installing
-      the final package.
-     </para>
```
adrianschroeter, line 89: oh, it was there already :/

xml/art-obs-beginners-guide.xml
sknorr: I think @tomschr wrote this from the perspective of OBS allowing to create many types of packages/images, but RPM still being the most important of those. As such, it does make sense to me. I also would not overburden the Beginner's Guide with too many theoretical concerns that do not affect RPM.

(I have to say that on the whole, this PR does not make much sense to me in its current form -- the wording updates it makes generally seem to make it less specific/useful.)

libnoon: Thanks a lot for your review, it is much appreciated.
On my side, I do deb packages and I'm interested in OBS to perform this task, so I'm actually trying to learn from its documentation; however certain things do not map correctly for deb packages and that's the main reason I'm proposing changes.  I'm trying to make it more universally correct while not complicating the reading.  My goal is that the next person in my case will have a less hard time.
I'm happy to take any advice to make it better.

tomschr: Maybe it would make sense to create two Beginners Guides: one for RPM base (openSUSE) and one for DEB (Debian)?
It also depends probably on how much text we could share between the two.

libnoon:
Here is my second try.  Could you please review it?

xml/art-obs-beginners-guide.xml
lrupp: > Maybe it would make sense to create two Beginners Guides: one for RPM base (openSUSE) and one for DEB (Debian)?
> It also depends probably on how much text we could share between the two.

# Packaging

For openSUSE based RPMs, we usually push people to read the [english openSUSE wiki](https://en.opensuse.org/Portal:Packaging). This is our reference documentation. Especially as it allows anyone to quickly adjust the content, once new features (including package macros) end up in the openSUSE distribution. With this, I would recommend to exclude the openSUSE packaging part completely and instead refer to the wiki.

About [Debian](https://en.opensuse.org/openSUSE:Build_Service_Debian_builds) / [Fedora](https://en.opensuse.org/Fedora_packaging) - or any other packaging: IMHO we should only refer to the specific pages in the openSUSE wiki. Things are really changing quickly - and people usually tend to not send PRs to Github, but edit the wiki instead. 

# Administration

A big difference are IMHO the configuration settings for the different distributions. 

These settings have to be done by the administrators of the Build Service. Here the wiki currently lacks this documentation, which is no surprise, as [OBS](https://build.opensuse.org) usually/magically (thanks to Adrian) provides them. But this is also what makes the live of any OBS admin hard: he has to figure out "how Adrian did this"? :-)

xml/art-obs-beginners-guide.xml
```diff
@@ -63,38 +63,38 @@ https://www.suse.com/communities/blog/suse-studio-integration/
...
+      </para>
       <para>
-       To successfully install and remove a package and all its contents,
-       the package manager needs to know which files and directories belong to
-       which package.
+        The installed package may require other packages to function
+        properly.  They are called the <emphasis>installation
+        requirements</emphasis>.
       </para>
      </formalpara>
```
tomschr, line 99: Same here. You need to move your end tag `</formalpara>` from line 99 after the `</para>` from line 93.
(GitHub doesn't allow me to add suggestions with deleted lines.)

tomschr (changes requested):
@libnoon I found your validation problems. See below. :slightly_smiling_face:

PR #199, opened 2021-07-02
document EL0 cpu flag

adrianschroeter:


xml/obs_build_job_constraints.xml
```diff
@@ -112,6 +112,13 @@ Constraint: linux:version:max 4.0</screen>
      <para> Example for project configuration:</para>
      <screen>Constraint: hardware:cpu:flag mmx
 Constraint: hardware:cpu:flag sse2</screen>
+     <para>A special flag is 'EL0', which can get added on specific armv8l systems.
+	   It is telling that it the hardware is supporting only execption level 0. This
+	   basically means no vm or kernel in 32bit is supported, but user land is
+	   supported. This flag can be used to block builds on such hardware, if
+	   no 64bit kernel is available for certain projects: </para>
```
sknorr, line 119: Rewording suggestion -- note that I was not sure about the following:
1. Is this an Arm-specific flag? (I reworded in a way that implies that it is currently only used for Arm but may be applicable more generally.
2. I hope I did not misunderstand "no vm or kernel in 32bit is supported".

```suggestion
     <para>
      <literal>EL0</literal is a special flag that that can be used on
      hardware that only supports level-0 exceptions, such as certain armv8l
      systems.
      This means that VM or 32-bit kernels are not supported but userland is
      supported.
      This flag can be used to block builds on such hardware if no 64-bit
      kernel is available for a project.
     </para>
```

PR #206, opened 2021-11-18
Document workflow runs

rubhanazeem:


xml/obs_scm_ci_workflow_integration.xml
```diff
@@ -778,6 +778,25 @@ workflow:
       </figure>
     </sect2>
 
+    <sect2 xml:id="sec.obs.obs_scm_ci_workflow_integration.setup.workflow_runs">
+      <title>Workflow Runs</title>
+
+      <para>For every SCM event, OBS record the request, and the response sent back to the SCM.
+      These records are stored in OBS in the form of Workflow runs. OBS will show details of every
+      event received from SCM. If an error occures in any of the workflow steps, the
+      workflow run will record the same response as it was sent to the SCM.</para>
```
hennevogel, line 787: Sounds like 2 times the same sentence. Maybe apart from *what* you should explain *why* we are doing this and *how* this is useful to the people reading this.

hennevogel (changes requested):

xml/obs_scm_ci_workflow_integration.xml
rubhanazeem: I've pushed some updates. Please let me know if you have more suggestions. :)

xml/obs_scm_ci_workflow_integration.xml
```diff
@@ -778,6 +778,25 @@ workflow:
       </figure>
     </sect2>
 
+    <sect2 xml:id="sec.obs.obs_scm_ci_workflow_integration.setup.workflow_runs">
+      <title>Workflow Runs</title>
+
+      <para>For every SCM event, OBS records the request and the response sent back to the SCM.
+      These records are stored in OBS in the form of Workflow runs. The main reason for these
```
dmarcoux, line 785:
```suggestion
      These records are stored in OBS in the form of workflow runs.
```

xml/obs_scm_ci_workflow_integration.xml
```diff
@@ -778,6 +778,25 @@ workflow:
...
+      workflow runs is to help the users to debug their workflows. If an error occurs in any of
```
dmarcoux, line 786:
```suggestion
      This helps with debugging workflows. If an error occurs in any of
```

dmarcoux (changes requested):

dmarcoux (approved):

eduardoj (approved):

PR #211, opened 2021-12-10
Fix YAML syntax in SCM/CI Workflow docs

perlpunk:
I found some mistakes on https://openbuildservice.org/help/manuals/obs-user-guide/cha.obs.scm_ci_workflow_integration.html


eduardoj (approved):
Thanks!

PR #218, opened 2022-01-26
Document paths in configure_repositories for SCM/CI integration

dmarcoux:
This isn't live yet, so this should be merged and deployed right before we deploy those changes. This way, users of the `configure_repositories` step know what needs to change when this breaking change is introduced.

To review the changes locally:
1. `docker-compose run --rm obs-docu daps -vv -d DC-obs-all html`
2. `firefox/other_browser build/obs-all/html/obs-all/cha.obs.scm_ci_workflow_integration.html`

Preview of the changes:
obs-docu-configure_repositories
obs-docu-configure_repositories-2

### DIFF
--- a/xml/obs_scm_ci_workflow_integration.xml
+++ b/xml/obs_scm_ci_workflow_integration.xml
@@ -492,13 +492,8 @@ rebuild_master:
             </listitem>
 
             <listitem>
-              <para>a target project, e.g.:
-              <emphasis>openSUSE:Factory</emphasis></para>
-            </listitem>
-
-            <listitem>
-              <para>a target repository, e.g.:
-              <emphasis>snapshot</emphasis></para>
+              <para>a list of paths, each having a target project (e.g: <emphasis>openSUSE:Factory</emphasis>)
+                and target repository (e.g: <emphasis>snapshot</emphasis>)</para>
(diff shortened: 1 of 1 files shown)


hennevogel (approved):

perlpunk:
Does that mean existing workflow configs won't be working anymore as soon as this goes live? Or is there a phase of deprecation for the old configuration?

dmarcoux:
@perlpunk This will be a breaking change. So as soon as the change goes live, `configure_repositories` step will have to be updated. We try as much as possible to avoid this, but sometimes it has to happen. It allows us to move faster while the SCM/CI integration is still part of the beta program. Sorry for the inconvenience.

While there isn't a deprecation phase for this, an error message in the workflow runs will link to the `configure_repositories` step documentation to help everyone update their `configure_repositories` step. Error messages are displayed on the workflow runs page in the web UI.

rubhanazeem (approved):

dmarcoux:
I've pushed changes to add missing section IDs. See the last commit.

rubhanazeem (approved):

dmarcoux:
@hennevogel had another look already. So I'll merge this.

PR #223, opened 2022-03-15
Use github actions to deploy obs-docs on obs-landing

krauselukas:
This GitHub action automatically builds the obs documentation and pushes the newly generated doc books to the obs-landing repository using a GitHub deployment key (basically an ssh key, that has only permission for a single repository).
It basically does the same that we did manually in the past https://github.com/openSUSE/obs-docu#update-documentation
See test runs on my fork https://github.com/krauselukas/obs-docu/actions/runs/1986616461

TODO:
- [x] Ignore files that do not influence the documentation itself, in order to prevent unnecessary commits on the obs-landing repository

### DIFF
--- a/.github/workflows/deploy-to-obs-landing.yaml
+++ b/.github/workflows/deploy-to-obs-landing.yaml
@@ -1,0 +1,63 @@
+name: deploy-to-obs-landing
+on:
+  push:
+    branches:
+      - master
+    paths-ignore:
+      - '.github/workflows/**'
+      - '!.github/workflows/deploy-to-obs-landing.yaml'
+      - '.travis.yml'
+      - 'Dockerfile'
+      - 'Dockerfile.base'
+      - 'README.md'
+      - 'docker-compose.override.yml.example'
+      - 'docker-compose.yml'
+jobs:
+  build-obs-docs:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Setup SSH Client
+        uses: webfactory/ssh-agent@v0.5.4
+        with:
+          ssh-private-key: ${{ secrets.DEPLOY_KEY_PRIVATE }}
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          ssh-key: ${{ secrets.DEPLOY_KEY_PRIVATE }}
+      - name: Checkout obs-landing repo
+        uses: actions/checkout@v3
+        with:
+          repository: openSUSE/obs-landing
+          path: obs-landing
+          ssh-key: ${{ secrets.DEPLOY_KEY_PRIVATE }}
+      - name: build
+        uses: addnab/docker-run-action@v3
+        with:
+          image: openbuildservice/obs-docu:latest
+          options: -v ${{ github.workspace }}:/github/workspace/
+          run: |
+            pushd /github/workspace/
+            for doc in obs-admin-guide obs-user-guide; do
+              source ./DC-$doc || exit 1
+
+              daps --verbosity=3 html || exit 1
+              daps --verbosity=3 epub || exit 1
+              daps --verbosity=3 pdf || exit 1
+
+              rm -rf obs-landing/help/manuals/${doc} || exit 1
+              # update html
+              cp -aL build/${doc}/html/${ROOTID} obs-landing/help/manuals/${doc} || exit 1
+
+              # update epub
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-to-obs-landing.yaml
```diff
@@ -0,0 +1,55 @@
...
+          options: -v ${{ github.workspace }}:/github/workspace/
+          run: |
+            pushd /github/workspace/
+            for doc in obs-admin-guide obs-user-guide; do
+              ROOTID=${doc%-guide}
+              source ./DC-$doc || exit 1
+
+              # update html
+              daps --verbosity=3 html || exit 1
+              daps --verbosity=3 epub || exit 1
```
tomschr, line 37: According to our statistics, EPUB isn't really used anymore. So maybe it would be a good opportunity to just don't offer this format?

.github/workflows/deploy-to-obs-landing.yaml
```diff
@@ -0,0 +1,55 @@
...
+          ssh-key: ${{ secrets.DEPLOY_KEY_PRIVATE }}
+      - name: build
+        uses: addnab/docker-run-action@v3
+        with:
+          image: openbuildservice/obs-docu:latest
+          options: -v ${{ github.workspace }}:/github/workspace/
+          run: |
+            pushd /github/workspace/
+            for doc in obs-admin-guide obs-user-guide; do
+              ROOTID=${doc%-guide}
```
tomschr, line 32: Sorry, but this isn't a good idea to set the `ROOTID` like this. This parameter is defined inside the DC file or not at all. It points to the ID of the root element. Maybe set it in your DC file.

No need to explicitly set it here, so I would recommend to remove it.

```suggestion
```

tomschr (commented):
@krauselukas Looks good Lukas! :+1: Just some minor comments. Hope it will help. :slightly_smiling_face:

.github/workflows/deploy-to-obs-landing.yaml
krauselukas: @hennevogel can we remove the EPUB format?

.github/workflows/deploy-to-obs-landing.yaml
krauselukas: @tomschr thanks for the feedback, I removed it, I didnt saw it is defined in the DC files :)

.github/workflows/deploy-to-obs-landing.yaml
krauselukas: @tomschr even its no in heavily used anymore, we thought it doesn't really hurt to still provide the EPUB version

krauselukas:
PR is ready, I will define the deployment keys on the repository and then this is ready to roll from my side :)

krauselukas:
> PR is ready, I will define the deployment keys on the repository and then this is ready to roll from my side :)

Done

.github/workflows/deploy-to-obs-landing.yaml
```diff
@@ -0,0 +1,63 @@
...
+              # update epub
+              mv build/${doc}/${ROOTID}_en.epub obs-landing/files/manuals/${doc}.epub || exit 1
+
+              # update pdf
+              mv build/${doc}/${ROOTID}_color_en.pdf obs-landing/files/manuals/${doc}.pdf || exit 1
+            done
+      - name: Commit files
+        run: |
+          pushd obs-landing
+          git config --local user.email "lkrause@suse.com"
```
dmarcoux, line 60: Could we use the email of the Build Solutions team here?

.github/workflows/deploy-to-obs-landing.yaml
tomschr: It's up to you, of course. :slightly_smiling_face: Just wanted to tell you, it's not very popular among our readers. :wink:

.github/workflows/deploy-to-obs-landing.yaml
krauselukas: sure, good point

krauselukas: @dmarcoux done

dmarcoux (approved):
I think it's possible to replace `addnab/docker-run-action@v3` by official instructions to run only certain steps in a Docker container. It doesn't need to happen now though.

PR #227, opened 2022-03-23
Clarify how event filters have to be defined

dmarcoux:
We already got feedback from users thinking that they could pass an array of events in the event filter.

To review the changes locally:
1. `docker-compose run --rm obs-docu daps -vv -d DC-obs-all html`
2. `firefox/other_browser build/obs-all/html/obs-all/cha.obs.scm_ci_workflow_integration.html`

Preview of the changes:
preview


dmarcoux:
[Clarify that event filter only accepts a single event (S)](https://trello.com/c/6HYVQqau/1859-clarify-that-event-filter-only-accepts-a-single-event-s)

saraycp (approved):

PR #232, opened 2022-03-28
Also commit new files in the deploy job

hennevogel:
git commit -a will only add known files.


hennevogel:
You can see what happened here

https://github.com/openSUSE/obs-docu/runs/5695401022?check_suite_focus=true

rubhanazeem (approved):
Tested locally and documentation is generated without validation errors.

PR #233, opened 2022-03-29
Add a use case with push events for tags

dmarcoux:
To test this locally:
1. Pull the changes from this PR, then run `daps2docker .`
2. Open the generated user guide in your favorite browser: `firefox /tmp/daps2docker-SXo2tVAc/obs-user-guide/html/book.obs-user/cha.obs.scm_ci_workflow_integration.html` (the string after `daps2-docker-` will be different).

Preview of the changes:
preview-1
preview-2

### DIFF
--- a/xml/obs_scm_ci_workflow_integration.xml
+++ b/xml/obs_scm_ci_workflow_integration.xml
@@ -1353,5 +1353,115 @@ workflow:
             architecture: x86_64</screen>
       <para></para>
     </sect2>
+
+    <sect2 xml:id="sec.obs.obs_scm_ci_workflow_integration.use_cases.create_package_releases_with_tags">
+      <title>Create Package on OBS for Every Software Release With Git Tags</title>
+
+      <para>You have a software project for which you mark releases with Git tags. For every release, you want to create
+      a package on OBS. This can be automated in a workflow with the <emphasis>branch_package</emphasis> step and the
+      <emphasis>tag_push</emphasis> event filter. Once the workflow is in place, every tag you push to your SCM repository
+      will branch a package on OBS and create, then build a package for the source code associated to the tag's commit.
+      This way, your users can always install a versioned release of your software project. You can also link one of those
+      versioned releases to another project on OBS if you need it as a dependency.</para>
+
+      <para>After the usual <link linkend="sec.obs.obs_scm_ci_workflow_integration.setup">setup</link> for
+      OBS workflows with tokens and webhooks, you will need:</para>
+
+      <itemizedlist>
+        <listitem>
+          <para>A package in OBS that you own, and for which you want to create releases. It will be the
+          <emphasis>source package</emphasis> (e.g.: <emphasis>home:jane/my_package</emphasis>) and it will
+          contain a <emphasis>_service</emphasis> file. When this package is branched by the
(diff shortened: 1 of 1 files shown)


dmarcoux:
[Add a use case with push events for tags (M)](https://trello.com/c/fVudlOMl/1858-add-a-use-case-with-push-events-for-tags-m)

saraycp (approved):

xml/obs_scm_ci_workflow_integration.xml
```diff
@@ -1353,5 +1353,128 @@ workflow:
...
+        source_project: home:jane
+        source_package: my_package
+        target_project: home:jane:releases
+  filters:
+    event: tag_push</screen>
+
+      <note>
+        <para>A Git tag is a pointer to a commit, and commits exist independently of branches. This is why
+        filtering on branches is <emphasis>not</emphasis> possible when working with push events for tags.</para>
+      </note>
```
hennevogel, line 1399: Why mention this?

hennevogel (commented):
Overall I'm missing a bit and explanation what will happen if I do all of this....

xml/obs_scm_ci_workflow_integration.xml
dmarcoux: I'm expecting questions about this. Should we remove this note?

dmarcoux: I think it won't be really common, so I'll remove it.

dmarcoux:
I've pushed again, please have another look.

saraycp (approved):
I think the first paragraph is more understandable now.
The rest, being more concise, are easier to read as well.

hennevogel:
I'm still missing description of the outcome of this, all of this seems like a very general technical description to me.

This is basically the only sentence we provide for our customers to decide if this is the *right* thing for them:

> Once the workflow is in place, every tag you push to your SCM repository will create a package on OBS for the source code associated to the tag's commit.

Where will this package be created? How will it look like? How can my packages users benefit from this? Is there something else I need to do?

hennevogel:
> Please have another look

Maybe I have missed something but I don't see any changes?

dmarcoux:
Here's the [latest diff](https://github.com/openSUSE/obs-docu/compare/3bcdf6701f0767f4e49ce4e8d3d01d05eb846307..9ed22f88c51f363cc2ca36f71eeb989ea0bcb23e).

With the current changes, I'm explaining why someone would want to set this up, while also giving details on how this works. I've removed a lot of extra and superfluous information from the text I wrote when I first submitted this PR. I believe that I have answered your previous questions: `Where will this package be created? How will it look like? How can my packages users benefit from this? Is there something else I need to do?`.

danidoni (approved):

danidoni:
> Here's the [latest diff](https://github.com/openSUSE/obs-docu/compare/3bcdf6701f0767f4e49ce4e8d3d01d05eb846307..9ed22f88c51f363cc2ca36f71eeb989ea0bcb23e).
> 
> With the current changes, I'm explaining why someone would want to set this up, while also giving details on how this works. I've removed a lot of extra and superfluous information from the text I wrote when I first submitted this PR. I believe that I have answered your previous questions: `Where will this package be created? How will it look like? How can my packages users benefit from this? Is there something else I need to do?`.

LGTM!

PR #248, opened 2022-10-11
Minor grammatical improvements

der-gabe:
Just a few things I noticed while reading the docs…

### DIFF
--- a/xml/obs_request_and_review_system.xml
+++ b/xml/obs_request_and_review_system.xml
@@ -173,8 +173,9 @@
     <title>Manually added reviews</title>
     <para> Reviewers can be added manually by anyone involved in a request.
      This can be used to hand over a review. In that situation the new reviewer
-     needs to be added and the old review needs to be accepted. The request
-     becomes declined when any of the reviewers are declining the request.
+     needs to be added and the original reviewer's own review needs to be accepted.
+     The request becomes declined when any of the reviewers are declining the
+     request.
     </para>
    </sect3>
    <sect3>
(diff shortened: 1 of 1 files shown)


xml/obs_request_and_review_system.xml
```diff
@@ -170,26 +170,26 @@
     certain package. Also, new maintainers of a package will see requests in
     case the old maintainer did not handle them. </para>
    <sect3>
-    <title>Manual added reviews</title>
+    <title>Manually added reviews</title>
     <para> Reviewers can be added manually by anyone involved in a request.
      This can be used to hand over a review. In that situation the new reviewer
-     needs to be added and the own review needs to be accept. The request
+     needs to be added and the old review needs to be accepted. The request
```
hennevogel, line 176: This is meant to say *own*. One user adds a review for another user, then accepts his *own* review to be out of the loop.

hennevogel (changes requested):

hennevogel:
Thank you for the contribution @der-gabe 💐

xml/obs_request_and_review_system.xml
der-gabe: Ah, I see now. Then it's a just a tiny bit ungrammatical… how about this:
```suggestion
     needs to be added and the original reviewer's own review needs to be accepted. The request
```
(This suggestion makes the line overly long, so don't apply it directly, but if you agree I will make the change in a separate commit and re-wrap the lines.)

der-gabe: Sorry for the delay, I had overlooked the :+1:

Issue #122, opened 2019-01-24
Examples about projconfig constraints is wrong


coolo:
xml/obs_build_job_constraints.xml says `unit="G" 4`

I followed this example and noticed it doesn't work. Then I checked the source and realized it needs to be unit=G

PR #263, opened 2023-02-05
#122 update xml/obs_build_job_constraints.xml 

JatinKumar001:
change unit="G" 4 to unit=G


hennevogel:
Why would you change this?

PR #265, opened 2023-02-09
Document ways to set configuration file location in workflows

hellcp-work:
Documents the way to configure custom configuration file locations in obs workflows, to let people know they can do that.


xml/obs_scm_ci_workflow_integration.xml
```diff
@@ -363,7 +363,8 @@
 
       <para>To do so, in the root directory of your GitHub repository, create
       a directory <emphasis>.obs</emphasis> which contains a file called
-      <emphasis>workflows.yml</emphasis>.</para>
+      <emphasis>workflows.yml</emphasis>. If you can't use that directory, you
```
saraycp, line 366: Nitpicking: 
 ```suggestion
      <emphasis>workflows.yml</emphasis>. If you don't want to use that directory, you
```

File: xml/obs_scm_ci_workflow_integration.xml [hunk 2]
```diff
@@ -401,6 +403,39 @@ rebuild_master:
       <para>OBS assumes the configuration file is placed in your target branch
       and not in other branches or forks.</para>
```
saraycp, line 404: Sorry, I didn't realize this before. I'd also remove this line, it's not 100% true anymore, and this way we only rely on the new Configure File Location section for all the related information.

```suggestion
```

saraycp (approved):

PR #276, opened 2023-03-30
add docu for keepfilerequires and ignoreconflicts

adrianschroeter:


xml/obs_build_config.xml
```diff
@@ -348,6 +348,20 @@ Keep: debbuild</screen>
...
+        <term><parameter>ignoreconflicts</parameter></term>
+        <listitem>
+                <para>Ignore defined conflicts of packages. By default these are repored as unresolvable.
+                      </para>
+        </listitem>
+       </varlistentry>
+       <varlistentry>
+        <term><parameter>keepfilerequires</parameter></term>
+        <listitem>
+                <para>Depencies to filenames get ignored by default. This switch can be used
```
eduardoj, line 361:
```suggestion
                <para>Dependencies to filenames get ignored by default. This switch can be used
```

eduardoj (dismissed):

xml/obs_build_config.xml
```diff
@@ -348,6 +348,20 @@ Keep: debbuild</screen>
                       add an exclamation mark (<literal>!</literal>) as prefix.</para>
         </listitem>
        </varlistentry>
+       <varlistentry>
+        <term><parameter>ignoreconflicts</parameter></term>
+        <listitem>
+                <para>Ignore defined conflicts of packages. By default these are repored as unresolvable.
```
eduardoj, line 354:
```suggestion
                <para>Ignore defined conflicts of packages. By default these are reported as unresolvable.
```

xml/obs_build_config.xml
```diff
@@ -348,6 +348,20 @@ Keep: debbuild</screen>
...
+        <term><parameter>ignoreconflicts</parameter></term>
+        <listitem>
+                <para>Ignore defined conflicts of packages. By default these are reported as unresolvable.
+                      </para>
+        </listitem>
+       </varlistentry>
+       <varlistentry>
+        <term><parameter>keepfilerequires</parameter></term>
+        <listitem>
+                <para>Dependencies to filenames get ignored by default. This switch can be used
```
Vogtinator, line 361: This is wrong in two ways:

- Direct file dependencies are never ignored, only indirect ones
- If FileProvides are specified, they are not ignored for indirect ones either

I suggest something like

> Dependencies on files are only fulfilled if matching FileProvides are specified in the project configuration. If those are missing, the dependency results in an "unresolvable" state for directly required files or in silent breaking of the dependency for indirectly required files. With this option, all file requires are honoured by default and lead to "unresolvable" if there are no matching FileProvides defined.

adrianschroeter: File requires of "Requires:" tag are ignored by default, only BuildRequires: to files are not.

But I am fine with your text as well, will change it.

xml/obs_build_config.xml
```diff
@@ -348,6 +348,20 @@ Keep: debbuild</screen>
                       add an exclamation mark (<literal>!</literal>) as prefix.</para>
         </listitem>
        </varlistentry>
+       <varlistentry>
+        <term><parameter>ignoreconflicts</parameter></term>
+        <listitem>
+                <para>Ignore defined conflicts of packages. By default these are reported as unresolvable.
```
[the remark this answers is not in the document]
adrianschroeter: that depends on the recepie format and is too much here IMHO

xml/obs_build_config.xml
Vogtinator: I prefer "direct" and "indirect" because it also makes sense outside of RPM, e.g. in kiwi or Dockerfile contexts

xml/obs_build_config.xml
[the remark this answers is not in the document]
Vogtinator: Currently I don't understand what it means, a reference to any other documentation would be really useful here.

xml/obs_build_config.xml
adrianschroeter: there you have actually another dimenson, direct and indirect packages for build env and notinstalled package sets.

xml/obs_build_config.xml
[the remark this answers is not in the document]
Vogtinator: I still don't understand what "defined" means here, but with the "image building tool" usecase I get how it could be used at least.

PR #278, opened 2023-04-12
Remove unused .png and .svg images from docs

saraycp:
Daps offers a command to detect which images are not used in a DC file:  `daps -d PATH_TO_DC_FILE list-srcfiles-unused --imgonly`.

As we have three DC files in the same directory (`DC-obs-all`, `DC-obs-admin-guide`,  `DC-obs-user-guide`), we run the check on the DC file of a _set_, `DC-obs-all`, which contains all articles and books in the project directory. Following [this guide](https://opensuse.github.io/daps/doc/cha.daps.user.img.html#sec.daps.user.imgs.manage).

Unused files with extensions `.dia` and `.dot` are not removed for the moment.


images/src/png/obs_lkw.png
krauselukas, line 1: @saraycp some reference to this image is commented in the books https://github.com/openSUSE/obs-docu/blob/67c410f4d03d9a4279be168a3a395c4d86742b56/xml/obs_best_practice_webui_usage.xml#L267 maybe removing it as well?

krauselukas: or maybe it got commented by accident 
image
guess the lkw icon used to show up next to the text. But I think its not a big loss to simply remove it, the text is clear enough without it IMO :)

saraycp: Oops, as I merged the other PR, this was merged too. The other PR was based on this one.
So I'll tackle the image you mention on a new one.

PR #279, opened 2023-04-12
Optimize .png images from docs

saraycp:
While generating the documents with Daps, we sometimes get warnings about images that are not optimized. Daps offers a command to optimize the _.png_ images used in the docs: `daps -d PATH_TO_DC_FILE optipng`.

After running it, more than 50 images are optimized (included in this PR) .


images/src/png/obs_lkw.png
krauselukas, line 1: @saraycp some reference to this image is commented in the books https://github.com/openSUSE/obs-docu/blob/67c410f4d03d9a4279be168a3a395c4d86742b56/xml/obs_best_practice_webui_usage.xml#L267 maybe removing it as well?

krauselukas: or maybe it got commented by accident 
image
guess the lkw icon used to show up next to the text. But I think its not a big loss to simply remove it, the text is clear enough without it IMO :)

danidoni (approved):

PR #280, opened 2023-04-17
Remove commented XML that added inline image

saraycp:
Follow-up  of #278

The image used in the commented XML was removed on a previous commit, as Daps considered it was not in use.
The XML has been commented on for 6 years now, it's safe to remove it.


saraycp:
@krauselukas, please :eyes:

krauselukas (approved):

PR #281, opened 2023-04-24
Add container-compression-format BuildFlag

adrianschroeter:


xml/obs_build_config.xml
```diff
@@ -305,6 +305,15 @@ Keep: debbuild</screen>
         </para>
        </listitem>
       </varlistentry>
+      <varlistentry>
+        <term><parameter>container-compression-format:FORMAT</parameter></term>
+       <listitem>
+        <para>
+          Sets a compression format for container layers. Possible values are gzip, zstd,
+          zstd:chunk. The usage of values beside gzip is only supported by podman and buildah.
```
hennevogel, line 313:
```suggestion
          zstd:chunk. Every value other than gzip is only supported by podman and buildah.
```

adrianschroeter: changed

PR #288, opened 2023-06-05
Add a link into a new chapter for the revamped API docs

danidoni:
The Glossary lacks a term explaining what an API is, and while we're at it, add a link to our new API Docs site.

image

### DIFF
--- a/xml/obs_glossary.xml
+++ b/xml/obs_glossary.xml
@@ -34,6 +34,20 @@
    </para>
   </glossdef>
  </glossentry>
+ <glossentry xml:id="obs.glos.apidoc">
+  <glossterm>API</glossterm>
+  <glossdef>
+   <para>
+    API stands for application programming interface. It let your product or service communicate with other products and services without having to know how they’re implemented.
+   </para>
+   <para>
+    The OBS API is located here: <link xlink:href="https://api.opensuse.org" />.
+   </para>
+   <para>
+    The documentation for the API is located here: <link xlink:href="https://api.opensuse.org/apidocs" />.
+   </para>
+  </glossdef>
+ </glossentry>
  <glossentry xml:id="obs.glos.appliance">
   <glossterm>Appliance</glossterm>
   <glossdef>
--- a/xml/obs_glossary.xml
+++ b/xml/obs_glossary.xml
@@ -38,7 +38,7 @@
   <glossterm>API</glossterm>
   <glossdef>
    <para>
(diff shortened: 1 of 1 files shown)


xml/obs_glossary.xml
```diff
@@ -34,6 +34,20 @@
    </para>
   </glossdef>
  </glossentry>
+ <glossentry xml:id="obs.glos.apidoc">
+  <glossterm>API</glossterm>
+  <glossdef>
+   <para>
+    API stands for application programming interface. It let your product or service communicate with other products and services without having to know how they’re implemented.
```
eduardoj, line 41:
```suggestion
    API stands for application programming interface. It lets your product or service communicate with other products and services without having to know how they’re implemented.
```
Small typo.

xml/obs_glossary.xml
```diff
@@ -34,6 +34,20 @@
...
+   </para>
+   <para>
+    The OBS API is located here: <link xlink:href="https://api.opensuse.org" />.
+   </para>
+   <para>
+    The documentation for the API is located here: <link xlink:href="https://api.opensuse.org/apidocs" />.
+   </para>
```
eduardoj, line 48:
```suggestion
   <para>
    The OBS API is documented here: <link xlink:href="https://api.opensuse.org/apidocs/" />.
   </para>
```
Providing the direct link to the OBS API doesn't add as much value as providing the link to the documentation, IMHO. Suggesting both links sounds to me kind of redundant. What do you think?

danidoni: Well, the Glossary talks about the API, not the documentation. Placing a link to the API documentation out of the blue, it's kind of disconnected. So to connect the documentation with the API I need to mention the API first.

eduardoj: Fair point. :+1:

tomschr: In my naive view, the main page of `api.opensuse.org` would contain all the necessary links. :wink:  That would make it easier for this glossary item as you only have to link to one link.

eduardoj: > In my naive view, the main page of `api.opensuse.org` would contain all the necessary links. wink That would make it easier for this glossary item as you only have to link to one link.

Noted down. We track now your suggestion in one of our cards: https://trello.com/c/oWhGOnN4/2975-change-what-to-show-in-https-apiopensuseorg

 Thanks for the feedback. Please keep it comming!

PR #293, opened 2023-07-19
Add table to document which event on the SCM will trigger which actions on OBS

rubhanazeem:
We document which SCM events we support and we also document what each step does for the different SCM events.

### DIFF
--- a/xml/obs_scm_ci_workflow_integration.xml
+++ b/xml/obs_scm_ci_workflow_integration.xml
@@ -1163,6 +1163,90 @@ test_build:
     </sect2>
   </sect1>
 
+  <sect1 xml:id="sec.obs.obs_scm_ci_workflow_integration.setup.event_action_and_status_report_table">
+    <title>Workflow Steps reference table</title>
+    <para>This table shows what will be returned to the SCM on different SCM events.</para>
+
+    <table>
+      <title>Workflow Steps reference table</title>
+      <tgroup align="left" cols="7">
+        <thead>
+          <row>
+            <entry>SCM events and actions</entry>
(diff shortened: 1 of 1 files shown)


xml/obs_scm_ci_workflow_integration.xml
```diff
@@ -983,6 +983,90 @@ test_build:
       </note>
     </sect2>
 
+    <sect2 xml:id="sec.obs.obs_scm_ci_workflow_integration.setup.event_action_and_status_report_table">
```
hennevogel, line 986: I think this would better fit into it's own section (`sect1`). I would put it between `SCM/CI Workflow Integration Setup` and `SCM/CI Workflow Versions`

xml/obs_scm_ci_workflow_integration.xml
```diff
@@ -983,6 +983,90 @@ test_build:
...
+      <title>SCM Events, Workflow Steps, and Status Report</title>
```
hennevogel, line 987: I would call this: `Workflow Steps reference table` and I think you need at least one sentence to explain what this is. Like: `This table shows which event on the SCM will trigger which actions on OBS.`

hennevogel (changes requested):
This is not only relevant to reporting. I think this is rather a complete new section in

xml/obs_scm_ci_workflow_integration.xml
rubhanazeem: Okay, I've added a short explanation for this table

hennevogel (approved):
Reviewed by commit

Issue #299, opened 2023-08-04
caches downloaded packages should be under `/var/tmp/oscbuild-packagecache`


bishal7679:
## Issue
caches downloaded packages shoudld be under `/var/tmp/oscbuild-packagecache` instead of `/var/tmp/osbuild-packagecache`

Page link :- https://openbuildservice.org/help/manuals/obs-user-guide/art.obs.bg#sec.obsbg.req

section :- **Hardware Requirements**

eduardoj:
Fixed by: #300.

PR #300, opened 2023-08-04
OBS caches download location is changed to `/var/tmp/oscbuild-packagecache`

bishal7679:
## Fixes :- #299


eduardoj (approved):
Thanks for the fix! And keep them coming!

eduardoj:
Hm... it is not a but, it is a feature. That is the defined default in osc. See https://github.com/openSUSE/osc/blob/b36c2101ba42c3a9b698ea67c853f6d0bca1e53c/osc/conf.py#L91

Reverting the change...

bishal7679:
> Hm... it is not a bug, it is a feature. That is the defined default in osc. See https://github.com/openSUSE/osc/blob/b36c2101ba42c3a9b698ea67c853f6d0bca1e53c/osc/conf.py#L91
> 
> Reverting the change...

Yes! Absolutely

PR #305, opened 2023-09-22
osc: Add documentation for cafile/capath

ziQ:
This PR adds documentation for `cafile`/`capath` options in `.oscrc`


ziQ:
Related to openSUSE/osc#1403

PR #308, opened 2023-10-24
Add the mention of supported event actions

hellcp-work:
Taken from https://github.com/openSUSE/open-build-service/blob/729c0721e9903f1824aa8e33dedd5396c6ba0a9c/src/api/app/models/scm_webhook.rb#L10-L11

### DIFF
--- a/xml/obs_scm_ci_workflow_integration.xml
+++ b/xml/obs_scm_ci_workflow_integration.xml
@@ -187,7 +187,7 @@
 
         <itemizedlist>
           <listitem>
-            <para>Pull requests</para>
+            <para>Pull requests and Merge requests</para>
           </listitem>
 
           <listitem>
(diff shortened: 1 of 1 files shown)


xml/obs_scm_ci_workflow_integration.xml
```diff
@@ -199,6 +199,58 @@
           </listitem>
         </itemizedlist>
 
+        <para>In addition, the Pull request events contain an field called
```
danidoni, line 202:
```suggestion
        <para>In addition, the Pull request events contains a field called
```

xml/obs_scm_ci_workflow_integration.xml
```diff
@@ -199,6 +199,58 @@
...
+          <listitem>
+            <para>synchronize</para>
+          </listitem>
+
+          <listitem>
+            <para>synchronized</para>
+          </listitem>
+        </itemizedlist>
+
+        <para>For Gitlab's Merge request event, the following actions are
```
danidoni, line 229:
```suggestion
        <para>For Gitlab's Merge request events, the following actions are
```
To be coherent with the previous paragraph

danidoni (approved):

saraycp (changes requested):

xml/obs_scm_ci_workflow_integration.xml
```diff
@@ -199,6 +199,58 @@
...
+          <listitem>
+            <para>reopen</para>
+          </listitem>
+
+          <listitem>
+            <para>update</para>
+          </listitem>
+        </itemizedlist>
+
         <para>Other SCMs like GitLab have similar behaviour; the main
```
saraycp, line 254: The paragraph which starts in this line should be moved right after the events list. 
Take into account we only list GitHub events and use this paragraph to invite users to discover the events from other SCMs.

Unlike the actions, that we list all of them (GitHub, GitLab, Gitea).

saraycp: An alternative would be listing all the possible events for all the SCMs. And then rephrase that paragraph to simply invite the users to read more (and don't move it).

saraycp (changes requested):

Issue #291, opened 2023-06-14
Mention statistics in the constraints chapter


hennevogel:
The [Build Job Constraints chapter](https://openbuildservice.org/help/manuals/obs-user-guide/cha.obs.build_job_constraints) does not mention the _statistics file / UI that helps you to figure out *which* constraints to use.

PR #323, opened 2024-02-08
user: rewrite parts of the Build Constraints chapter

smithfarm:
Fixes: https://github.com/openSUSE/obs-docu/issues/291

### DIFF
--- a/xml/obs_build_constraints.xml
+++ b/xml/obs_build_constraints.xml
@@ -3,43 +3,125 @@
   <!ENTITY % entities SYSTEM "entity-decl.ent">
     %entities;
 ]>
-<chapter version="5.1" xml:id="cha.obs.build_job_constraints"
+<chapter version="5.1" xml:id="cha.obs.build_constraints"
  xmlns="http://docbook.org/ns/docbook"
  xmlns:xi="http://www.w3.org/2001/XInclude"
  xmlns:xlink="http://www.w3.org/1999/xlink" >
- <title>Build Job Constraints</title>
+ <title>Build Constraints</title>
  <info/>
- <para> Build job constraints can define requirements for the hardware or
-  software of the build host. Constraints can be defined per package or for
-  repositories. </para>
- <para> The build constraints for an entire project or specific repositories is
-  part of the project config. For each constraint, it contains a line</para>
-  <screen>Constraint: &lt;SELECTOR&gt; &lt;STRING&gt;</screen> 
-  <para>The selector is a colon-separated list. </para>
- <para> The build constraints for a package are part of the package sources, as
-  a _constraints XML source file (validated on submission). The _constraints
-  source file can contain the values listed below. </para>
- <para>In addition to that some build recipe parsers support to parse constraints
-  also from the build descriptions, eg. in spec files. These can also be used
-  conditionally in their context. For example to apply these only on some architectures
-  or only for specific multibuild flavors.</para>
- <para>Constraints defined in project build config, _constraints and the build recipe are merged
-  in this order. This means a constraint for the same topic overwrites the one from the former location.</para>
- <para>NOTE: If no build host meets the constraints, the package will turn to a failed state.</para>
+ <para> Using build constraints, one can define hardware and/or software
+  requirements that the Build Service will then use to narrow down the pool
+  of build workers that are qualified to build a particular project, package or
+  repository.
+ </para>
+ <para> By setting appropriate build constraints, one can avoid situations
+  in which builds fail due to insufficient memory or disk space, wrong kernel
+  version, etc. </para>
  <sect1>
-  <title><literal>hostlabel</literal></title>
-  <para> The hostlabel is any string which can be assigned to build hosts when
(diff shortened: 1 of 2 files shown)


xml/obs_build_constraints.xml
```diff
@@ -3,43 +3,125 @@
...
-   in another instance is usually not possible. Use any of the other
-   constraints if possible. </para>
-  <para> Example for _constraints file:</para>
-   <screen>&lt;constraints exclude="false"&gt;
-  &lt;hostlabel&gt;benchmark_runner&lt;/hostlabel&gt;
+  <title>Setting constraints</title>
+  <para> Constraints can be set for a particular package by including a
+   _constraints file in the package. </para>
+  <para> Constraints can also be included directly in the build recipe (spec,
+   dsc, etc.), provided the build recipe parser supports it. </para>
```
hennevogel, line 25: It would be nice if you would name the ones we *know* support it right now...

xml/obs_build_constraints.xml
```diff
@@ -3,43 +3,125 @@
...
+  <warning>
+   <para>It's not clear how the reader would find out if a given build recipe
+    parser supports BuildConstraints and it would even be nice here to link to a
+    discussion of what, specifically, is meant by "build recipe". </para>
+  </warning>
```
smithfarm, line 30: ^^^ @hennevogel

adrianschroeter: no, please don't write such a thing in the documentation. The documentation is here to explain that.

smithfarm: Please note, this is just a draft and these warnings will disappear before the "draft" flag is removed. Just to be clear.

xml/obs_build_constraints.xml
adrianschroeter: it is written there already: "Example for rpm spec file or Dockerfile:"

adrianschroeter:
On Freitag, 9. Februar 2024, 10:13:39 CET Nathan Cutler wrote:
> @smithfarm commented on this pull request.
> 
> 
> 
> > -   starting the bs_worker process. It can be used to run on specific hosts,
> -   which may be used for running benchmarks in a reproducible way. This
> -   constraint can also be defined as a negative definition using the
> -   <literal>exclude=true</literal>
> -   attribute. However, the hostlabel is always specific to one OBS instance.
> -   You should avoid it as much as possible, since building with this constraint
> -   in another instance is usually not possible. Use any of the other
> -   constraints if possible. </para>
> -  <para> Example for _constraints file:</para>
> -   <screen>&lt;constraints exclude="false"&gt;
> -  &lt;hostlabel&gt;benchmark_runner&lt;/hostlabel&gt;
> +  <title>Setting constraints</title>
> +  <para> Constraints can be set for a particular package by including a
> +   _constraints file in the package. </para>
> +  <para> Constraints can also be included directly in the build recipe (spec,
> +   dsc, etc.), provided the build recipe parser supports it. </para>
> 
> I can write expicitly that only spec and Dockerfile are supported, but then the moment support is expanded to another recipe that statement will be wrong...

yes, that is our reality that we always need to run after the documentation
on implementation :)

-- 

Adrian Schroeter ***@***.***>
Build Infrastructure Project Manager

SUSE Software Solutions Germany GmbH, Frankenstraße 146, 90461 Nürnberg, Germany
(HRB 36809, AG Nürnberg) Geschäftsführer: Ivo Totev

hennevogel:
> yes, that is our reality that we always need to run after the documentation on implementation :)

Documentation is part of the implementation...

smithfarm:
> yes, that is our reality that we always need to run after the documentation on implementation :)

@adrianschroeter I understand the sentiment and that this is the current reality, but it doesn't *have* to be that way. I can imagine an API call `/buildconstraints/supported/buildrecipeparsers` which would return the real answer to the question "which recipe parsers support buildconstraints?" And then the documentation could tell folks how to get the real answer, instead of presenting a false answer.

adrianschroeter:
On Freitag, 9. Februar 2024, 13:09:56 CET Nathan Cutler wrote:
> > yes, that is our reality that we always need to run after the documentation on implementation :)
> 
> @adrianschroeter I understand the sentiment and that this is the current reality, but it doesn't *have* to be that way. I can imagine an API call `/buildconstraints/supported/buildrecipeparsers` which would return the real answer to the question "which recipe parsers support buildconstraints?" And then the documentation could tell folks how to get the real answer, instead of presenting a false answer.

This is not an api feature, but a build process tooling support task.

And even when you have that, you would still need to document 
the specific syntax for each build description format.

-- 

Adrian Schroeter ***@***.***>
Build Infrastructure Project Manager

SUSE Software Solutions Germany GmbH, Frankenstraße 146, 90461 Nürnberg, Germany
(HRB 36809, AG Nürnberg) Geschäftsführer: Ivo Totev

smithfarm:
OK, this discussion has helped me and I now know how to deal with this "supported build recipes" problem. Wait for it ;-)

smithfarm:
will open a clean PR

PR #327, opened 2024-02-22
Document Report and Moderation process

saraycp:
Instructions about how users can report misconduct and moderators can act accordingly.

I attach a PDF to make it easier to review: 

[Moderation User Guide.pdf](https://github.com/openSUSE/obs-docu/files/14403848/Moderation.User.Guide.pdf) (updated at 26/02/2024 12:29 CET).


xml/obs_moderation.xml
```diff
@@ -0,0 +1,208 @@
...
+  xmlns="http://docbook.org/ns/docbook"
+  xmlns:xi="http://www.w3.org/2001/XInclude"
+  xmlns:xlink="http://www.w3.org/1999/xlink" >
+  <title>Moderation</title>
+  <info/>
+ 
+  <para>As a platform where a lot of social interaction happens, OBS is not free of spam,
+    harmful content or other outcomes of user misconduct.
+    Therefore, OBS provides features which help any user report problematic content and help moderators to act accordingly.
+    Those features are described below, and they are helpful for admins, staff members, moderators and users.
```
danidoni, line 16: Shouldn't those be different `<para>`?

saraycp: No, I intend to have one paragraph with three lines. That's how it looks in the resulting document.

I separated the lines in the code to make it easier to read.

danidoni: Alright

xml/obs_moderation.xml
```diff
@@ -0,0 +1,297 @@
...
+          </imageobject>
+        </mediaobject>
+      </figure>
+    </sect2>
+
+    <sect2>
+      <title>How To Moderate?</title>
+      <para>
+        As a moderator you can click on the yellow warning about reports, usually displayed close to the element.
+        There, you can read all the reports and make a <emphasis>decision</emphasis>. Write the reason why you agree, <emphasis>Favor</emphasis>, or disagree, <emphasis>Cleared</emphasis>.
```
danidoni, line 189:
```suggestion
        There, you can read all the reports and make a <emphasis>decision</emphasis>. Write the reason why you agree (Favor), or disagree (Cleared).
```
I think writing it between parentheses makes it easier to convey that "Favor" and "Cleared" are synonyms, otherwise you could understand that there are _four_ reasons....

danidoni, line 189:
```suggestion
        There, you can read all the reports and make a <emphasis>Decision</emphasis>. Write the reason why you agree, <emphasis>Favor</emphasis>, or disagree, <emphasis>Cleared</emphasis>.
```
Also, decision should be capitalized, like we do with Favor and Cleared, because those are concepts we use across the UI (part of our "shared language", like talking about Packages and Projects or BsRequests).

tomschr: As an alternative you could also create a glossary entry and link to it. That would make it crystal clear and is independent of any spelling.

tomschr: As an alternative you could also create a glossary entry and link to it. That would make it crystal clear and is independent of any spelling.

saraycp: @tomschr Done!
Could you please have a look at the new glossary entry and the cross-reference, please?

tomschr: @saraycp Looks good! :+1: Thanks for adding it (I just thought, it was a crazy idea. :laughing:)

saraycp: It's a good idea and we could probably add more words related to the most recent features. This is a good start!

Issue #298, opened 2023-07-31
SCM CI says that gitea does not support scoped API tokens - but it does


dcermak:
In https://openbuildservice.org/help/manuals/obs-user-guide/cha.obs.scm_ci_workflow_integration.html#sec.obs.obs_scm_ci_workflow_integration.setup.token_authentication.how_to_authenticate_obs_with_scm it is said that gitea does not support scoped API tokens. However that is no longer true as of gitea 1.19:
![image](https://github.com/openSUSE/obs-docu/assets/45594031/b8d64a4d-9fce-4e3f-b3c5-dfbe8ad43e30)

It would be nice if the documentation would be updated accordingly listing the required scopes.

smithfarm:
What are the required scopes? I'm guessing only the `repo` scope is needed and the rest not?

PR #333, opened 2024-03-25
user/obs_scm_ci_workflow_integration: gitea now supports token scopes

smithfarm:
Fixes: https://github.com/openSUSE/obs-docu/issues/298


dcermak:
> @dcermak Does this work for you?

I have never used gitea with the OBS SCM CI integration, so I don't know, sorry

smithfarm:
OK, but Gitea's token scopes are known (or at least assumed) to be modeled after Github's.

PR #341, opened 2024-05-27
Add "keepobsolete" PublishFlag

adrianschroeter:


xml/obs_build_config.xml
```diff
@@ -644,6 +644,20 @@ Support: pax debbuild</screen>
...
+        <term><parameter>keepobsolete</parameter> (OBS 2.11 or later)</term>
+        <listitem>
+           <para>
+             The default behaviour of OBS is to remove binaries when a package object gets removed,
+             even when the project is publish disabled (but the package may have been enabled).
+           </para>
+	   <para>
+             This flag is changing this behaviour to keep binaries in published repositories
+             even when a package gets removed. As consequence this flag has only an effect on
+             publish disabled projects.
```
smithfarm, line 657: I do not understand how "this flag only has an effect on publish disabled projects" is a consequence here. From the description, it sounds like the flag should have an effect on all projects, regardless of whether publishing is enabled or not.

xml/obs_build_config.xml
```diff
@@ -644,6 +644,20 @@ Support: pax debbuild</screen>
            </para>
         </listitem>
        </varlistentry>
+       <varlistentry>
+        <term><parameter>keepobsolete</parameter> (OBS 2.11 or later)</term>
+        <listitem>
+           <para>
+             The default behaviour of OBS is to remove binaries when a package object gets removed,
```
smithfarm, line 651: instead of just "remove binaries", would it make sense to write "remove binaries from published repositories" (because OBS also offers a way to get at non-published binaries via "osc getbinaries")?

Even though I suppose "osc getbinaries" will not work after the package gets removed...

xml/obs_build_config.xml
adrianschroeter: well, if the project has publish enabled the binaries get removed not matter if you have this flag or not.

The surprising thing for the user is that even when you have publish disabled on project level, the package removal operation has an effect by default, but all other operations have not.

This flag is changing this behaviour.

xml/obs_build_config.xml
adrianschroeter: we could .... however, it removes binaries everywhere, including published repositories. But also from build area, so it is not limited to published repos.... therefore I wrote this in this generic words.

xml/obs_build_config.xml
smithfarm: Current default behavior (no keepobsolete): when a package is deleted, any published binaries will be deleted, regardless of publish setting.

With keepobsolete: "I have a project with some published binaries. Then I disable publishing on the project. Then I delete a package (which has some old binaries published). The binaries are not deleted, because keepobsolete is set. If publishing is enabled, the published binaries are deleted along with the package, as usual."

adrianschroeter: that is correct, but we can not take this text 1:1 into the docu using "I", right?

xml/obs_build_config.xml
smithfarm: Ah, in that case, we should write "to remove all binaries".

smithfarm (approved):
I think I understand it now, and I'm having difficulty thinking of a better way to phrase it ;-)

PR #355, opened 2024-07-03
obs_build_config: refrain from nesting screen inside para

smithfarm:
Nesting screen inside para is a violation of the GeekoDoc XML schema and will cause the CI build to fail if/when we (eventually) switch to GeekoDoc.


hennevogel:
So daps switched the default from DocBook5 to GeekoDoc. This switches it back.

hennevogel (approved):

PR #369, opened 2024-07-31
obs_build_constraints.xml: drop double-quotes from unit= values

smithfarm:
Fixes: https://github.com/openSUSE/obs-docu/issues/122

### DIFF
--- a/xml/obs_build_constraints.xml
+++ b/xml/obs_build_constraints.xml
@@ -198,7 +198,7 @@ Constraint: hardware:disk:size unit=M 4000
    When specifying constraints using XML syntax, attribute values must be
    enclosed in double-quotes (<literal>unit="G"</literal>,
    <literal>exclude="true"</literal>, etc.), while in project configurations and
-   build recipes the attributes must be given <emphasis>without</emphasis>
+   build recipes the values must be given <emphasis>without</emphasis>
    quotes (<literal>unit=G</literal>, <literal>exclude=true</literal>).
    </para></note>
   <sect2>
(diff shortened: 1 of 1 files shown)


xml/obs_build_constraints.xml
```diff
@@ -188,6 +188,19 @@ Constraint: hardware:disk:size unit=M 4000
...
+   via a <literal>_constraints</literal> file, XML syntax is used, while a
+   different syntax is used when specifying constraints in a project
+   configuration or a build recipe (e.g., an RPM spec file). In this section,
+   both syntaxes are described for each qualifier. </para>
+  <note><para>
+   When specifying constraints using XML syntax, attribute values must be
+   enclosed in double-quotes (<literal>unit="G"</literal>,
+   <literal>exclude="true"</literal>, etc.), while in project configurations and
+   build recipes the attributes must be given <emphasis>without</emphasis>
+   quotes (<literal>unit=G</literal>, <literal>exclude=true</literal>).
```
hennevogel, line 202: This sounds like a bug that should be fixed instead of described to me...

mlschroe (approved):

xml/obs_build_constraints.xml
smithfarm: The follow-up question, I suppose, would be: "a bug in what?"

If we say it's a bug in the build constraints parser that it does not tolerate double-quotes around values (that cannot contain spaces), then we must also say it's a bug in XML that it requires double-quotes around values, even when those values do not include spaces?

hennevogel: the parser is something we write, the XML standard not :)

PR #380, opened 2024-08-30
art-obs-beginners-guide.xml: address terminology confusion

smithfarm:
The word "project" has a special meaning in the OBS context, so let's be careful when using it in the OBS documentation.


hennevogel (approved):

hennevogel:
Changes like this are priceless 💚  Thank you!

PR #398, opened 2025-02-12
Clarify behaviour of configure_repositories step

danidoni:


xml/obs_scm_ci_workflow_integration.xml
```diff
@@ -716,6 +716,23 @@ rebuild_master:
...
+              <para><emphasis>pull request opened</emphasis> will configure a number of repositories and architectures for the target project explained before.</para>
+	    </listitem>
+	    <listitem>
+              <para><emphasis>rebasing an existing pull request</emphasis> will configure any new repositories and architectures coming from the workflows file and leave the old ones untouched.</para>
+	    </listitem>
+	    <listitem>
+              <para><emphasis>closing an existing pull request</emphasis> will destroy the project, so any repositories and architectures will be gone too.</para>
+	    </listitem>
+	    <listitem>
+              <para><emphasis>reopening a pull request</emphasis> will configure the repositories and architectures like it is a newly opened pull_request.</para>
```
hennevogel, line 732: All of this is not true for a `pull_request` event with any of the mentioned actions *if* the only step is `configure_repository`.

hennevogel (changes requested):

xml/obs_scm_ci_workflow_integration.xml
danidoni: So you think we have to clarify that we need an associated `pull_request` step or this won't work?

hennevogel: This always works. You need to clarify what it does to the mentioned scenarios.

- to existing repositories/architectures that are part of the workflow.yml
- to existing repositories/architectures that are not part of the workflow.yml

hennevogel: What about all the other steps and what about if it's the only step? I think you are trying to explain different workflow, instead of describing what the step does...

danidoni: But what the step does is to configure repositories and architectures for the target project... And what the target project is depends on the received event and the previous step...

hennevogel: The only difference is the name of the project it does things too. How is this relevant?

danidoni: Because when we receive a pull request event we use a specific convention for the project name and when we get a push event we use another convention.

danidoni: Let's see if the last change is better

hennevogel (approved):

PR #403, opened 2025-02-25
Add Release tag to prjconf chapter

hennevogel:


smithfarm (changes requested):
It *seems* that the keywords/parameters/tags are listed here in alphabetical order. At least, they should be.

So please move the new "Release" keyword/parameter/tag down so it appears between "RegistryURL" and "Repotype"?

hennevogel:
@smithfarm done

smithfarm (approved):

PR #406, opened 2025-03-25
Change mv to cp to avoid a permission denied error

rubhanazeem:
Currently `build-obs-docs` job fails on `build html` step with the error `mv: cannot move '/tmp/daps2docker-*/...' to '../obs-landing/help/manuals/obs-admin-guide': Permission denied`. 

This pr attempts to resolve this error.


hennevogel:
you could also just `cp` instead of `mv`...

rubhanazeem:
> you could also just `cp` instead of `mv`...

Good idea. Let me change it to `cp` instead of changing permissions

hennevogel (approved):

PR #422, opened 2025-07-04
Drop Hermes from svg

bmwiedemann:
Drop Hermes from svg
because it is gone for a long time
and people are using this diagram in their slides.


bmwiedemann:
Do we want to mention rabbitmq?

images/src/svg/obs-concept.svg
```diff
@@ -1,345 +1,1417 @@
...
+      
+      
+      
+      
+      
+      
+      
+      
+      
+
```
eduardoj, line 105:
```suggestion
      
```

Are all these empty lines needed?

hennevogel:
Where are we even using this SVG?

PR #460, opened 2026-07-06
Documentation for orphan branches

saraycp:
After this implementation: https://github.com/openSUSE/open-build-service/pull/19925

### DIFF
--- a/xml/obs_scm_ci_workflow_integration.xml
+++ b/xml/obs_scm_ci_workflow_integration.xml
@@ -498,6 +498,16 @@ rebuild_master:
         </figure>
       </sect3>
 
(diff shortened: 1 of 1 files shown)


hennevogel:
I don't think this paragraph is needed. It's just confusing, you say don't do something for something that does not happen anymore after https://github.com/openSUSE/open-build-service/pull/19925

saraycp:
> I don't think this paragraph is needed. It's just confusing, you say don't do something for something that does not happen anymore after [openSUSE/open-build-service#19925](https://github.com/openSUSE/open-build-service/pull/19925)

Ok! We'll need to inform on our blog post for users that are suffering from "fake yaml files"

hennevogel:
> Ok! We'll need to inform on our blog post for users that are suffering from "fake yaml files"

I only know of openQA that did this...

